PR #21, opened 2015-05-28
Example tag models

markstephens:
Example tag models
- an example of how the json payload will look when the tracking requests are sent through.

### DIFF
--- a/resources/tag_models/event.json
+++ b/resources/tag_models/event.json
@@ -1,0 +1,29 @@
+{
+  "tag": {
+	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
+	"type": "event", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
+  },
+
+  "event": {
+	"page_id": "", // ID of the page on which this event happened.
+	"parent_id": "", // Parent event ID if this is a child event.
+	"category": "", // Main category of the event, likely often the component name e.g. video
+	"action": "", // The action performed e.g. play
+	"key": "", // defines what value is passed
+
+	"value": "value",
+	"data": { // Data is an object and can have several variables. These values do not get extracted into separate columns in Redshift, so can only be accessed through inefficient Redshift functions: http://docs.aws.amazon.com/redshift/latest/dg/json-functions.html
+	}
+
+  },
+
+  "user": {
+	"session_id": "", // User's current logged in session ID, from membership
+	"guid": "", // User ID on ft.com
+	"passport_id": "", // User ID on ft.com - legacy
+	"spoor_id": "" // Our device ID, used to keep track of anon users.
+  }
+}
--- a/resources/tag_models/event.json
+++ b/resources/tag_models/event.json
@@ -13,11 +13,9 @@
 	"category": "", // Main category of the event, likely often the component name e.g. video
 	"action": "", // The action performed e.g. play
 	"key": "", // defines what value is passed
-
 	"value": "value",
 	"data": { // Data is an object and can have several variables. These values do not get extracted into separate columns in Redshift, so can only be accessed through inefficient Redshift functions: http://docs.aws.amazon.com/redshift/latest/dg/json-functions.html
 	}
-
   },
 
   "user": {
--- a/resources/tag_models/page.json
+++ b/resources/tag_models/page.json
@@ -1,0 +1,57 @@
+{
+  "tag": {
+	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
+	"type": "page", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
+  },
+
+  "page": {
+	"url": "",
+	"referrer": "",
+	"channel": "", // Our products e.g. ft.com, webapp, flipboard
+	"uuid": "", // UUID of the page, won't always get this
+	"asset_type": "section", // Is this available from an API during enrichment?
+	"page_class_1": "",
+	"page_class_2": "",
+	"sitemap": "Sections.Africa"
+  },
+
+  "device": {
+	"user_agent": "", // User agent, usually get this from headers, but are cases where tags are forwarded on and need to explicitly set the user’s agent.
+	"ip": "" // IP address, usually get this from headers, but are cases where tags are forwarded on and need to explicitly set the user’s IP
+  },
+
+  "user": {
+	"session_id": "", // User's current logged in session ID, from membership
+	"guid": "", // User ID on ft.com
+	"passport_id": "", // User ID on ft.com - legacy
+	"spoor_id": "" // Our device ID, used to keep track of anon users.
+  },
+
+  "other": { // Anything else specific to this rendering of the page
+	"ad_refresh": "",
+	"barrier": "",
+	"segment_id":"",
+	"offer_id": "",
+	"seg_id":"",
+	"dfp_zone": "",
+	"dfp_site": "",
+	"dfp_targeting": "",
+	"twitter_login": false,
+	"gift_article_used": "",
+	"step_number": "",
+	"step_name": "",
+	"payment_term": "",
+	"app_edition": "",
+	"transition_action": "",
+	"search_term": "",
+	"sub": "",
+	"ticker": "",
+	"brand": "",
+	"weekend": false,
+	"edition": "UK",
+	"theme": ""
+  }
+}
--- a/resources/tag_models/page.json
+++ b/resources/tag_models/page.json
@@ -10,12 +10,16 @@
   "page": {
 	"url": "",
 	"referrer": "",
-	"channel": "", // Our products e.g. ft.com, webapp, flipboard
+	"product": "", // Our products e.g. ft.com, webapp, flipboard
 	"uuid": "", // UUID of the page, won't always get this
-	"asset_type": "section", // Is this available from an API during enrichment?
-	"page_class_1": "",
-	"page_class_2": "",
-	"sitemap": "Sections.Africa"
+	"data": { // Anything else specific to this rendering of the page
+	  "barrier": "", // Whether the page had a barrier
+	  "segment_id":"", // Marketing Segment ID
+	  "offer_id": "", // Marketing Offer ID
+	  "seg_id":"", // Marketing SegID
+	  "step_number": "", // Step number within a defined process or funnel.
+	  "step_name": "" // Step name within a defined process or funnel.
+	}
   },
 
   "device": {
@@ -28,30 +32,5 @@
 	"guid": "", // User ID on ft.com
 	"passport_id": "", // User ID on ft.com - legacy
 	"spoor_id": "" // Our device ID, used to keep track of anon users.
-  },
-
-  "other": { // Anything else specific to this rendering of the page
-	"ad_refresh": "",
-	"barrier": "",
-	"segment_id":"",
-	"offer_id": "",
-	"seg_id":"",
-	"dfp_zone": "",
-	"dfp_site": "",
-	"dfp_targeting": "",
-	"twitter_login": false,
-	"gift_article_used": "",
-	"step_number": "",
-	"step_name": "",
-	"payment_term": "",
-	"app_edition": "",
-	"transition_action": "",
-	"search_term": "",
-	"sub": "",
-	"ticker": "",
-	"brand": "",
-	"weekend": false,
-	"edition": "UK",
-	"theme": ""
   }
 }


AlbertoElias:
Must the key names be those? Some aren't easy to understand without the comment

markstephens:
@AlbertoElias which keys? Most of them aren't fixed at all, and can be changed.

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
+	"type": "event", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
+  },
+
+  "event": {
+	"page_id": "", // ID of the page on which this event happened.
+	"parent_id": "", // Parent event ID if this is a child event.
+	"category": "", // Main category of the event, likely often the component name e.g. video
```
AlbertoElias: We need to clarify if this will be category or component /cc @triblondon

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+	"data": { // Data is an object and can have several variables. These values do not get extracted into separate columns in Redshift, so can only be accessed through inefficient Redshift functions: http://docs.aws.amazon.com/redshift/latest/dg/json-functions.html
+	}
+
+  },
+
+  "user": {
+	"session_id": "", // User's current logged in session ID, from membership
+	"guid": "", // User ID on ft.com
+	"passport_id": "", // User ID on ft.com - legacy
+	"spoor_id": "" // Our device ID, used to keep track of anon users.
```
AlbertoElias: spoor seems like a weird name for this. why not `device_id`?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
+	"type": "page", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
+  },
+
+  "page": {
+	"url": "",
+	"referrer": "",
+	"channel": "", // Our products e.g. ft.com, webapp, flipboard
```
AlbertoElias: Maybe call it `product`?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"uuid": "", // UUID of the page, won't always get this
+	"asset_type": "section", // Is this available from an API during enrichment?
```
AlbertoElias: I don't understand this one

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"page_class_1": "",
```
AlbertoElias: Don't understand these two

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+  "device": {
+	"user_agent": "", // User agent, usually get this from headers, but are cases where tags are forwarded on and need to explicitly set the user’s agent.
+	"ip": "" // IP address, usually get this from headers, but are cases where tags are forwarded on and need to explicitly set the user’s IP
+  },
+
+  "user": {
+	"session_id": "", // User's current logged in session ID, from membership
+	"guid": "", // User ID on ft.com
+	"passport_id": "", // User ID on ft.com - legacy
+	"spoor_id": "" // Our device ID, used to keep track of anon users.
```
AlbertoElias: Same as above

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+
+  "other": { // Anything else specific to this rendering of the page
+	"ad_refresh": "",
+	"barrier": "",
+	"segment_id":"",
+	"offer_id": "",
+	"seg_id":"",
```
AlbertoElias: What's seg?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"dfp_zone": "",
```
AlbertoElias: What's dfp?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"dfp_site": "",
+	"dfp_targeting": "",
+	"twitter_login": false,
+	"gift_article_used": "",
+	"step_number": "",
```
AlbertoElias: What do you mean by step?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"step_name": "",
+	"payment_term": "",
+	"app_edition": "",
+	"transition_action": "",
```
AlbertoElias: What's a transition action?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"search_term": "",
+	"sub": "",
```
AlbertoElias: What's sub?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"ticker": "",
```
AlbertoElias: What's ticker?

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
+{
+  "tag": {
+	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
```
commuterjoy: semver friendly?

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+	"type": "event", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
```
commuterjoy: perhaps just take the SQS message ID? or the MD5 of the message.

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
+  },
+
+  "event": {
+	"page_id": "", // ID of the page on which this event happened.
+	"parent_id": "", // Parent event ID if this is a child event.
```
commuterjoy: Do this on the server somewhere in the backend / redshift.

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+  "tag": {
+	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
+	"type": "event", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
+  },
+
+  "event": {
+	"page_id": "", // ID of the page on which this event happened.
```
commuterjoy: Perhaps replace with URL? Everyone understands URL.

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+	"parent_id": "", // Parent event ID if this is a child event.
+	"category": "", // Main category of the event, likely often the component name e.g. video
+	"action": "", // The action performed e.g. play
+	"key": "", // defines what value is passed
```
commuterjoy: would prefer just allowing some arbitrary JSON structure for the meta data. then people can model what they need in a sensible way.

resources/tag_models/event.json
```diff
@@ -0,0 +1,29 @@
...
+
+	"value": "value",
+	"data": { // Data is an object and can have several variables. These values do not get extracted into separate columns in Redshift, so can only be accessed through inefficient Redshift functions: http://docs.aws.amazon.com/redshift/latest/dg/json-functions.html
+	}
+
+  },
+
+  "user": {
+	"session_id": "", // User's current logged in session ID, from membership
```
commuterjoy: don't want any of this in the client code really. we can extract it from the cookies on the server.

resources/tag_models/page.json
commuterjoy: +1

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
+{
```
commuterjoy: still don't really see the merit in maintain two separate models. a page view is just a type of event that happens when the client decides a page has been seen.

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+  "tag": {
+	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
+	"version": "", // Version of the tracking client e.g. "JS v1.2"
+	"type": "page", // Type of request (page or event)
+	"id": "", // ID of this request, must be unique
+	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
```
commuterjoy: not sure what offset is for?

resources/tag_models/page.json
commuterjoy: We can derive anything like sections the from the content API.

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"dfp_site": "",
+	"dfp_targeting": "",
+	"twitter_login": false,
+	"gift_article_used": "",
+	"step_number": "",
+	"step_name": "",
+	"payment_term": "",
+	"app_edition": "",
+	"transition_action": "",
+	"search_term": "",
```
commuterjoy: it's just part of the URL?

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+	"sub": "",
+	"ticker": "",
+	"brand": "",
```
commuterjoy: content api

resources/tag_models/page.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+
+  "user": {
+	"session_id": "", // User's current logged in session ID, from membership
+	"guid": "", // User ID on ft.com
+	"passport_id": "", // User ID on ft.com - legacy
+	"spoor_id": "" // Our device ID, used to keep track of anon users.
+  },
+
+  "other": { // Anything else specific to this rendering of the page
```
commuterjoy: I think a lot of these are product specific (ft.com?) and would like to not include them in o-tracking.

resources/tag_models/page.json
markstephens: Yep, none of these are set in stone, just habit.

Rob Shilston just suggested we use cmdb name - although I think that'll end up being not the same definition.

resources/tag_models/page.json
lucas42: offset is used by the web app for tracking things which happen offline.  We don't want to trust the client's absolute time.  Instead we work out the difference in time between the event happening and it getting sent to the server.  That way we can subtract it from the time on the server and work out when the event took place.

resources/tag_models/event.json
markstephens: "category" as may not always be from a component.

resources/tag_models/page.json
markstephens: Yep

resources/tag_models/page.json
markstephens: we can get from CAPI.

resources/tag_models/page.json
markstephens: seg_id is a marketing ID, been around for years. It's just the name it is.

resources/tag_models/page.json
markstephens: These are step in a process, such as subscribing, might have 3 steps.
The process may span multiple pages, so they can't be hierarchical ids.

resources/tag_models/event.json
markstephens: in the real implementation it is the semver.
https://github.com/Financial-Times/o-tracking/blob/master/main.js#L47

resources/tag_models/event.json
markstephens: We need the ID created in the client to be able to attach child events later on on the page.

resources/tag_models/event.json
markstephens: You can't do this based on the events coming in in sequence, it has to be explicitly set ID's.

resources/tag_models/event.json
markstephens: It won't work. What if the user has multiple tabs with the same page?
Even if that wasn't an issue, the processing would still be relying on fuzzy matching based on timestamps which is naff.

resources/tag_models/event.json
markstephens: Bu then we can't model it in the database, we'd need 50 columns all with different types of value in, and then what if we get 51 values in a following event.

Most events are performing a single action.

resources/tag_models/event.json
markstephens: Mostly this will not exist, but not if the event comes from FlipBoard/Google Newstand, and they need to specify the session ID somewhere.

resources/tag_models/page.json
markstephens: Yes, assuming it's in the URL.

resources/tag_models/page.json
markstephens: Yep, CAPI.

resources/tag_models/page.json
markstephens: Thanks Luke, oTracking uses the same logic.

resources/tag_models/event.json
markstephens: "spoor" is the running name of this project.
It's not really a device_id, it symbolises a user on a device.

resources/tag_models/page.json
markstephens: Same answer as above ;-)

resources/tag_models/page.json
markstephens: I'm not fixed on this either, especially as they are very similar anyway.

The current thinking is we have a "page" at each end of the pipeline, why try to fit it into one, to expand it out again later.
The other idea is that a page has more values, whereas a event typically has one value.
Finally, a page is special when it comes to reporting, it is a different "type" of thing compared to an event.

lucas42: The web app has lots of edge cases which are tricky to decide whether they're pages or not.  Technically, the user only every loads a single page and everything else is javascript.  So we had to make many arbitrary decisions about which interactions counted as a new page (eg is opening an overlay a separate page?  What if that overylay takes up the entire screen? etc etc)

markstephens: We are more flexible in NGDA than we are with iJento, the main reason we have to identify pages in iJento is the ability to tell if they are "counted content". iJento could only do this on a page.

In WebApp you might only send one page for the whole app and then send events OR we might might model pages again. Partly it's a technical decision, partly it's a people-who-are-reporting-on-this-stuff decision.

In NGDA we are flexible to handle both, we are also thinking about a world without pages in the far future, but thats outside the scope of the NGDA project.

resources/tag_models/event.json
triblondon: Should clarify that this value is from a defined set of options, if that's the case.  'category' is otherwise a pretty ambiguous term.

resources/tag_models/event.json
triblondon: This is not public API though, is it?  This data is created and dispatched inside of oTracking, one assumes.

resources/tag_models/event.json
triblondon: This doesn't seem like public API territory. Is all this intended to be set as part of event payloads by the source code, rather than being added by the internals of o-tracking?

resources/tag_models/event.json
triblondon: As we discussed earlier today, Origami has a big issue with parent events.  We don't like it conceptually: the 'parent' is actually describing an event that happened at a different point in time, and no developer will understand that intuitively.  Origami wants to create technology that people understand how to use intuitively so you don't have to 'wrap your head around it' as a prerequisite to using it correctly.

Separately, implementation is complex, because it requires disparate parts of the page code to maintain shared state, which is an inherently bad pattern, and one that will also lead to memory leaks.

We proposed a system of 'collation ids' for an event, so that instead of passing a parent, we can associate multiple events together in a group, eg:

```
{
  "collation_ids": [
    "131bec90-055e-11e5-b939-0800200c9a66",
    "131c13a1-055e-11e5-b939-0800200c9a66"
  ]
}
```

resources/tag_models/event.json
triblondon: Don't think this should be public API.  But I'd also suggest using a name that's less prone to misunderstanding, eg 'page_session_id'

resources/tag_models/page.json
commuterjoy: @lucas42 yup. Whatever your arbitrary decisions about what constitutes a page view it's just a type of event though.

resources/tag_models/event.json
commuterjoy: @markstephens IIRC the use cases for needing this sort of thing weren't that strong?

resources/tag_models/event.json
commuterjoy: If you make a request to the collection end point you'll get an anonymous ID sent back. You can store that on your client and send it back every time that user makes a request.

`curl -si https://spoor-api.ft.com/px.gif | grep -a X-Anon`

... later I guess it will also be a Set-Cookie response.

commuterjoy:
BTW, this is what's collecting the data from o-tracking - https://spoor-docs.herokuapp.com/

resources/tag_models/event.json
commuterjoy: In that case I'd remove it and just let people use the additional meta data for their needs.

resources/tag_models/event.json
commuterjoy: Mmm, as discussed elsewhere, I'd just get rid of child events. If someone want to group the data by user, page, or date sequence, that's all possible later down the line.

markstephens: Why not do it in the client? It's easy to do it there and makes no real difference to the events being sent through to have an extra ID. 

We need child events, it is a business requirement. We can't just do counts of things it's too basic. 

Finally, how would you do it down stream? It's not possible to join them up accurately, the code would be hugely complex and it would be inefficient and slow. (Ref: multiple browser tabs, multiple components on a page and trying to fuzzy match based on dates)



> On 28 May 2015, at 22:35, Matt Chadburn <notifications@github.com> wrote:
> 
> In resources/tag_models/event.json:
> 
> > @@ -0,0 +1,29 @@
> > +{
> > +  "tag": {
> > +	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
> > +	"version": "", // Version of the tracking client e.g. "JS v1.2"
> > +	"type": "event", // Type of request (page or event)
> > +	"id": "", // ID of this request, must be unique
> Mmm, as discussed elsewhere, I'd just get rid of child events. If someone want to group the data by user, page, or date sequence, that's all possible later down the line.
> 
> —
> Reply to this email directly or view it on GitHub.
> 

-- 

------------------------------
This email was sent by a company owned by Pearson plc, registered office at 
80 Strand, London WC2R 0RL.  Registered in England and Wales with company 
number 53723.

resources/tag_models/event.json
markstephens: They are. We need the consequence of events, not just counts.

It is a business requirement. 

Regardless, I don't see why not have it. It makes no difference to the pipeline if there's an extra ID in there, and it provides far richer information. And allows the analytics team to do their job. 



> On 28 May 2015, at 22:17, Matt Chadburn <notifications@github.com> wrote:
> 
> In resources/tag_models/event.json:
> 
> > @@ -0,0 +1,29 @@
> > +{
> > +  "tag": {
> > +	"api_key": "", // String - API key - Make sure the request is from a valid client (idea nicked from Keen.io) useful if a page gets copied onto a Russian website and creates noise
> > +	"version": "", // Version of the tracking client e.g. "JS v1.2"
> > +	"type": "event", // Type of request (page or event)
> > +	"id": "", // ID of this request, must be unique
> > +	"offset": 0 // Delay of this tag between event happening and being sent to server - milliseconds
> > +  },
> > +
> > +  "event": {
> > +	"page_id": "", // ID of the page on which this event happened.
> > +	"parent_id": "", // Parent event ID if this is a child event.
> @markstephens IIRC the use cases for needing this sort of thing weren't that strong?
> 
> —
> Reply to this email directly or view it on GitHub.
> 

-- 

------------------------------
This email was sent by a company owned by Pearson plc, registered office at 
80 Strand, London WC2R 0RL.  Registered in England and Wales with company 
number 53723.

resources/tag_models/page.json
AlbertoElias: But what is it?

resources/tag_models/event.json
commuterjoy: >  it is a business requirement.

Can you remind everyone of the business requirements? The business is often wrong, quixotic, or tied to the past ways of doing things.

resources/tag_models/event.json
markstephens: @triblondon no it's not public API, it happens inside o-tracking.
Additionally, as Matt mentioned, the spoor api will set a cookie if the user doesn't CTM or is noscript.

resources/tag_models/event.json
markstephens: It's not public API. `page_id` is correct, it's the ID of a page.

resources/tag_models/event.json
markstephens: @triblondon 

The collation ID's were a nice idea, but we've thought about it further and we feel we can't do it.

To be able to store that in a database it would be a many to many relationship. e.g.

Event table

id | type | data
---- | ---- | ----
1 | page | { ... }
2 | event | { ... }
3 | event | { ... } 

Collation table

event_id | collation_id
----------- | ---------------
1 | 2
1 | 3

The most common query would be: "what events happened on a page?". To answer that question it has to do 2 joins, whereas with the "parent, child" hierarchy it only needs to do one. 2 joins is bad from a performance point of view, but probably also from an analytics person who's reporting on the data PoV as it's more complex.

The performance issue is made worse when you consider we estimate to have 5.5 billion events per year (more with the help of origami). We also estimate an event will have on average, 2 collation ID's, sometimes more, sometimes less. That makes the collation table at 11 billion rows per year.

We will want want to keep 3 years of data, which will make the collation table 30-40 billion rows long.

So, we think it's not going to be very performant doing many-to-many joins over that table and therefore prefer the hierarchy model. It's not easy to turn the collation model into the hierarchy model later on as how do you tell which is the correct or most important collation ID?


Also, I don't follow the argument it is confusing to developers, these aren't "javascript" (or programatic) events and child events (or any kind of inheritance), it's just another piece of data to supply with an event.

To avoid the memory leak issue, I believe the best idea it to create a separate o-tracking guid component which gens guids for components to use when sending events.

resources/tag_models/event.json
markstephens: We need to know the consequence of things.

Say you ping an event which is "content read".

We need to know why that content was read:
- Maybe it is because the user looked at an article (the page -> event hierarchy)
- Maybe it was because the recommended reads component loaded (event -> event hierarchy)
  - and then the user clicked on one of the reads (event -> page hierarchy)
- Maybe it was because a snippet loaded on fastFT and the user clicked read more (event -> event hierarchy)
- Maybe an A/B test changed something on the page and rendered a new component (page -> event -> event)

Without a hierarchy we know none of the above.

Then expand it out to all the different types of event: nav interaction, video interaction, slideshows, interactive graphics. We need to know what happened (counts of events), but also why (hierarchy).

commuterjoy: Just capture and store context around the event? If you have `<div data-trackable="foo">foo</div>` and that sort of thing then you can understand what's going on in a lot of detail.

resources/tag_models/event.json
triblondon: On understanding: Why don't you do a survey?  My vote is that your parent-child event model is extremely counter-intuitive. If everyone else thinks it makes perfect sense, then there's no problem here.  YOU understand your model extremely well.  The question is do you want others to understand it quickly and easily and therefore instinctively send good data without training, or do you want to battle with dodgy data coming in as a result of misunderstanding the model, and run regular training courses to teach people how the model works?

On implementation: the simple issue is that we want no shared state.  It's not loading o-tracking that I think will cause leaks, it's maintaining a ton of shared state everywhere so that components know what their 'parent event' is.  Any solution that avoids shared state is fine from a component development perspective.  So for example, if you want to organise events in your backend like this:

```
Page session
 \-- Video
  |-- Player 1
  \-- Player 2
    \-- Clip 1
      |-- Play
      \-- Scrub
```

Then to me, none of these are events except for Play and Scrub.  The others are all aggregations/collations of those events.  Now, if you accept events with a payload like this:

```
{
  collations: [
    "videoclip:12345",
    "o-video-instance:0000-0000-000-0000",
    "origami:o-video"
  ]
}
```

Then when that event arrives at o-tracking, you could add some higher level collations, like the page session ID, and then consider that sequence to be a hierarchy, if that is best for your model.

Incidentally, I'm no expert in data modelling and information retrieval, but I seem to remember something about sparse table algorithms that would potentially enable you to do this without a massive n:m join.  The most common collation types that you might want to query could be promoted to first-class event properties, like 'page session ID', since o-tracking is likely to add that to every single event.

If if you do opt to use a hierarchical model you should be able to convert it later into a flat collation system simply by treating every ancestor of an event as one of its collation IDs.

resources/tag_models/event.json
triblondon: > Without a hierarchy we know none of the above.

I think you guys have blinkers on for this hierarchy thing. I have been running the phrase "event -> event hierarchy" over in my head for about 5 minutes and I have not the slightest clue what it means.  I know it's frustrating to you, especially when you talk to lots of different people, so maybe it's worth having a larger gathering to discuss this point?

markstephens: But why not use an ID? It's more accurate and saves having to send through a lot of data with each request. WebApp, for example, won't be keen on unnecessarily large payloads.

In your suggestion, we'd have to send through the xpath of the element, plus repeat info about the page and other things? Why is that better than an id?

Also, you assume events will always be from a html page. Already we're talking about collecting email events. Opens and clicks are a good example of needing a hierarchical model too. A parent/child ID is the only consistent way.

resources/tag_models/event.json
markstephens: Understanding
I don't have time to do a survey, our current supplier could be bust in a month.
The normal use-case is still simple, just ping an event and don't worry about ID's.
There are specific use-cases where the relationship is needed, which will involve a defined statement of work, and user story, and testing, and assistance, and the rest of it.

Implementation
At least by having a guid generating component, only the component has to remember the "parent" id for the next event it sends - it's not shared state.

Algorithms
It's got to be achievable in SQL, there are a team of 30-40 people who know sql, plus it has to work with an off-the-shelf reporting tool which knows sql and how to do joins.

Finally
We do have to maintain the status quo too, millions in revenue are attached to this data, it's used everywhere - going into reports and more systems.
It's beyond the scope of this project to change every report, every business process, every third party system to accept a new model.

AlbertoElias: Mark, just a quick recap on why o-tracking can't convert the collation ids into a hierarchical model before pushing it into Spoor?

markstephens: Do you mean in the client?

In Andrew's example:
```
{
  collations: [
    "videoclip:12345",
    "o-video-instance:0000-0000-000-0000",
    "origami:o-video"
  ]
}
```

How do I programatically know which order these collation id's should go in? e.g. is instance a parent of videoclip and videoclip is a child of o-video? Or is it another way around?
I'd have to keep a map or something like that for each type of component/event.

AlbertoElias: You could consider the order in which the elements are added to the collations array to be the hierarchy structure, and the last value in the array to be the one assigned to the parent_id key /cc @triblondon

markstephens: I think thats getting more complex for a developer than remembering an ID, don't you think?

resources/tag_models/page.json
markstephens: Yep, agree, I'm going to make these the same format/structure, it's going to be easier for the enrichment processes to have things in the same place too.

resources/tag_models/event.json
triblondon: There are two threads here - one is our opinion on whether a hierarchical data model for analytics makes sense.  Alberto and I don't think it does, but if there's no flexibility on that or everyone else is in favour of heirarchy then fine.  In any case, we are willing to stop engaging with this point if it's proving unhelpful.

The other thread is the API used by developers.  As long as we do not have to maintain data between scopes that should have no need to communicate with one another, that will mitigate our strongest issue with the linking between events.  So to be clear, that means that no event triggered within a component will ever need to reference any event triggered outside of that component.

commuterjoy:
Hi - I've written some thoughts here - https://github.com/Financial-Times/o-tracking/pull/32

https://github.com/Financial-Times/o-tracking/blob/flat-model/docs/event.md

PR #22, opened 2015-05-28
Event formats and naming conventions.

markstephens:
- Update tests to avoid any left-over values from previous tests
- Set the "channel" correctly


test/event.js
```diff
@@ -1,16 +1,18 @@
...
 "use strict";
 
-var assert = require("assert"),
-	track_event = require("../src/javascript/event.js");
+var assert = require("assert");
 
 describe('event', function () {
 
-	var server, userID;
+	var server,
```
AlbertoElias: http://origami.ft.com/docs/syntax/js/#one-var-per-line

AlbertoElias:
lgtm

PR #26, opened 2015-06-02
Stop oTracking retaining values between tracking events.

markstephens:
Fairly major bug, the module was "remembering" values from the previous event in the next event!


AlbertoElias:
lgtm!

PR #30, opened 2015-06-05
Optimisations

markstephens:
- Switch to uuid generator instead of rolling own
- Cleanup unused functions
- Listen for CustomEvent detail object


commuterjoy:
:-1: I think all uuid generation should happen on the server. 

`curl -si spoor-api.ft.com/px.gif | grep Set-Cookie`

The client's job should be to accept that UUID and send it back next time.

markstephens:
Doing that, means the component won't fulfil the requirements.
It needs to set ID's in the browser.

commuterjoy:
why? the first time it talks to spoor it reads the uuid generates, stores it in a cookie. then read that from the cookie for all future requests.

markstephens:
We'd be relying that the spoor-api responds and the cookie is set before the next tracking event is sent.

There might be multiple events sent milliseconds apart, the first will be attached to the spoor-id but the immediately following events won't.

markstephens:
In noscript/fail CTM mode - it doesn't matter as there aren't any subsequent on-page events.

triblondon:
Agree with @commuterjoy.  The `node-uuid` dependency being included here in turn depends on node's `crypto`, which is browserifyable, but is enormous (Sam just built it, and it adds 3KB minified, 28KB unminified).

markstephens:
Ok, I'll revert back to the previous ID creation method I had then.

commuterjoy:
> We'd be relying that the spoor-api responds and the cookie is set
> before the next tracking event is sent.

Edge case? How many times has that happened in the last 6 months?

markstephens:
It happens every page impression on ft.com. The video player in the right rail announces it's loading.

This would be exaggerated in Next if we get every component to send a "I'm loading" event.

commuterjoy:
I doubt we'd want to be sending `i'm loading` events. If a component doesn't load then it's broken.

markstephens:
Depends if different components load for different users, or in different situations, then we'd want to know how it affects usage and interaction of the page/app/thing.

Regardless, it doesn't make sense to "hope" that the next request will be sent late enough. The implementation should be reliable. We don't know what events might be required in the future..

PR #32, opened 2015-06-10
Flat model

commuterjoy:
A proposal for more general model to represent page and events, and contains some thoughts on what I'd like to see the [Spoor API](https://spoor-docs.herokuapp.com) collect.

### DIFF
--- a/docs/event.md
+++ b/docs/event.md
@@ -9,20 +9,21 @@
 
 ## Spec 
 
-- `event.source` and `event.api-key` are MANDATORY, everything else is OPTIONAL
+`event.source` and `event.api-key` are MANDATORY, everything else is OPTIONAL
 
 This allows a low barrier to entry for anyone wanting to submit events to Spoor.
 
 ## Enrichments 
 
-There's a whole series of enrichments that we could do.
+Each payload will be scanned for the presence of these properties.
 
-Each payload will be scanned for the presence of these properties. If present the properties will be decorated and/or tokenised. 
+If present the properties will be decorated and/or tokenised. 
 
 To date we've discussed :-
 
 - If `user.session.token` is included it will be validated against the session api and expanded in to a membership uuid.
 - If `content.uuid` is included it will be expanded in to a set of content api meta-data.
+- ...
 
 ## Format
 
--- a/docs/event.md
+++ b/docs/event.md
@@ -70,6 +70,7 @@ Each payload will be scanned for the presence of these properties. If present th
 }
 ```
 
+```
 [1]  The general convention might be something like `product.sub-system.environment` - Eg, 'webapp.render.uat'
 [2]  Or in absence appended by Spoor. 
 [3]  So, the token becomes -> { session: { token: '...', uuid: '...', valid: true }
@@ -81,3 +82,4 @@ Each payload will be scanned for the presence of these properties. If present th
 [8]  Where the event happened. If not provided this is expanded by a call to Maxmind.
 [9]  An enrichment
 [10] Possibly not that useful?
+```
--- a/docs/event.md
+++ b/docs/event.md
@@ -3,8 +3,9 @@
 - It's JSON.
 - In this model there is no hierarchy (cf. o-tracking proposal), and only a couple of mandated fields.
 - If certain fields are present in the payload they will enriched by Spoor and appended to the payload.
-- I've tried to keep terminology human understable - Eg, the removal of 'categoy', 'page', 'key', 'value', 'data' should all be inherent in the data structure etc.
-- It is intentionally abstract and assumes the consumers of the event stream can manipulate the data in to their desired format.
+- I've tried to keep terminology human understable - Eg, the removal of 'categoy', 'page', 'key', 'value', 'data' should all be inherent in the data structure etc. Anything bespoke goes in 'meta'.
+- It is intentionally abstract and assumes the consumers of the event stream can manipulate the data in to their desired storage format.
+- It's for use by server and client-side systems sending events to Spoor.
 
 ## Spec 
 
@@ -12,14 +13,18 @@
 
 This allows a low barrier to entry for anyone wanting to submit events to Spoor.
 
-## Enchrichment 
+## Enrichments 
 
 There's a whole series of enrichments that we could do.
 
 Each payload will be scanned for the presence of these properties. If present the properties will be decorated and/or tokenised. 
 
+To date we've discussed :-
+
 - If `user.session.token` is included it will be validated against the session api and expanded in to a membership uuid.
 - If `content.uuid` is included it will be expanded in to a set of content api meta-data.
+
+## Format
 
 ```
 {
--- a/docs/event.md
+++ b/docs/event.md
@@ -1,7 +1,7 @@
 ## Notes
 
 - It's JSON.
-- In this model there is no hierarchy (cf. o-tracking), and only a couple of mandated fields.
+- In this model there is no hierarchy (cf. o-tracking proposal), and only a couple of mandated fields.
 - If certain fields are present in the payload they will enriched by Spoor and appended to the payload.
 - I've tried to keep terminology human understable - Eg, the removal of 'categoy', 'page', 'key', 'value', 'data' should all be inherent in the data structure etc.
 - It is intentionally abstract and assumes the consumers of the event stream can manipulate the data in to their desired format.
@@ -34,7 +34,24 @@ Each payload will be scanned for the presence of these properties. If present th
         "anonymous-id": "...",									// An anonymous token, supplied by the client [2] 
         "session": {
             "token": "asdf324dfag1ds%asdf1A-1sadsadf",  		// FT Session token - expanded in to a uuid and valid flag [3]
-        }
+        },
+		"device": {
+			"layout": "S",										// require('o-grid').getCurrentLayout 
+        	"orientation": "portrait"							// require('o-viewport').getOrientation
+		},
+		"geo": {												// [8]
+            "country": "MA",
+            "area_code": "0",
+            "longitude": "-5.000",
+            "latitude": "32.000",
+            "country_name": "Morocco",
+            "continent": "AF"
+		},
+		"ab": {
+			"test-name": "variant"								// [9]
+		},
+		"flags": [ ],											// Flags/Toggles the user has set [10]
+		"isStaff": true											// [9]
     },
     "content": {
         "uuid": "", 				// Content uuid [4] 
@@ -43,7 +60,7 @@ Each payload will be scanned for the presence of these properties. If present th
 		"type":	"content"			// Some sort of controlled list  
     },
 	"meta": {						// [7]
-		"anything": "goes",
+		"anything": "goes here",
 		"could: [ "be", "an", "array" ],
 		"or" : { "might": { "be": ["an", "object"] }}
 	},
@@ -53,12 +70,14 @@ Each payload will be scanned for the presence of these properties. If present th
 }
 ```
 
-[1] The general convention might be something like `product.sub-system.environment` - Eg, 'webapp.render.uat'
-[2] Or in absence appended by Spoor. 
-[3] So, the token becomes -> { session: { token: '...', uuid: '...', valid: true }
-[3] I will scan any Cookies in the header for a FTSession token as well.
-[4] Content will typically be a page, but could be a uuid for something else like a video. Will be expanded in to a CAPI v2 object -> capi: { "headline", "...", "published": "...", ... }
-[5] Expanded in to a URL object -> http://nodejs.org/api/url.html
-[6] I think I might also allow this in a `authorization` header.
-[7] This is intentionally vague.
-
+[1]  The general convention might be something like `product.sub-system.environment` - Eg, 'webapp.render.uat'
+[2]  Or in absence appended by Spoor. 
+[3]  So, the token becomes -> { session: { token: '...', uuid: '...', valid: true }
+[3]  I will scan any Cookies in the header for a FTSession token as well.
+[4]  Content will typically be a page, but could be a uuid for something else like a video. Will be expanded in to a CAPI v2 object -> capi: { "headline", "...", "published": "...", ... }
+[5]  Expanded in to a URL object -> http://nodejs.org/api/url.html
+[6]  I think I might also allow this in a `authorization` header.
+[7]  This is intentionally vague.
+[8]  Where the event happened. If not provided this is expanded by a call to Maxmind.
+[9]  An enrichment
+[10] Possibly not that useful?
--- a/docs/event.md
+++ b/docs/event.md
@@ -1,0 +1,64 @@
+## Notes
+
+- It's JSON.
+- In this model there is no hierarchy (cf. o-tracking), and only a couple of mandated fields.
+- If certain fields are present in the payload they will enriched by Spoor and appended to the payload.
+- I've tried to keep terminology human understable - Eg, the removal of 'categoy', 'page', 'key', 'value', 'data' should all be inherent in the data structure etc.
+- It is intentionally abstract and assumes the consumers of the event stream can manipulate the data in to their desired format.
+
+## Spec 
+
+- `event.source` and `event.api-key` are MANDATORY, everything else is OPTIONAL
+
+This allows a low barrier to entry for anyone wanting to submit events to Spoor.
+
+## Enchrichment 
+
+There's a whole series of enrichments that we could do.
+
+Each payload will be scanned for the presence of these properties. If present the properties will be decorated and/or tokenised. 
+
+- If `user.session.token` is included it will be validated against the session api and expanded in to a membership uuid.
+- If `content.uuid` is included it will be expanded in to a set of content api meta-data.
+
+```
+{
+    "event": {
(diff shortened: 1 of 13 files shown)


docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+    },
+    "user": {													// Who generate the event
+        "user-agent": "Mozilla ...",
+        "ip": "123.123.123.123",
+        "anonymous-id": "...",									// An anonymous token, supplied by the client [2] 
+        "session": {
+            "token": "asdf324dfag1ds%asdf1A-1sadsadf",  		// FT Session token - expanded in to a uuid and valid flag [3]
+        },
+		"device": {
+			"layout": "S",										// require('o-grid').getCurrentLayout
```
georgecrawford: This is only applicable to `o-grid` users. We'd need to provide proprietary flags here, but I guess that's acceptable.

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+```
+{
+    "event": {
+        "source": "email.simple-email-service",  				// Name of the sender's system [1]
+        "type": "opened",  										// The type of event that happened, Eg. 'pageview', 'dwell', 'play' etc.
+		"api-key":	"0f7464b4-3f4d-11e4-984b-00144feabdc0"		// Sender-specific key [6]
+    },
+    "user": {													// Who generate the event
+        "user-agent": "Mozilla ...",
+        "ip": "123.123.123.123",
```
georgecrawford: If you're offering to pull the user session token from the Cookie header, would the same be done for UA and IP if set, or do you want to mandate that these are explicitly set?

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+```
+[1]  The general convention might be something like `product.sub-system.environment` - Eg, 'webapp.render.uat'
+[2]  Or in absence appended by Spoor. 
+[3]  So, the token becomes -> { session: { token: '...', uuid: '...', valid: true }
+[3]  I will scan any Cookies in the header for a FTSession token as well.
+[4]  Content will typically be a page, but could be a uuid for something else like a video. Will be expanded in to a CAPI v2 object -> capi: { "headline", "...", "published": "...", ... }
+[5]  Expanded in to a URL object -> http://nodejs.org/api/url.html
+[6]  I think I might also allow this in a `authorization` header.
+[7]  This is intentionally vague. Allow anyone to store anything. Consumers own responsibility for filtering what they need.
+[8]  Where the event happened. If not provided this is expanded by a call to Maxmind.
```
georgecrawford: To be explicit, that would be expanded from the IP only, or also CDN headers if set?

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+[9]  An enrichment
+[10] Possibly not that useful?
```
georgecrawford: It might be to us. The client currently knows its 'QA' flags (from http://app.ft.com/qatool, our equivalent of http://next.ft.com/__toggler), but we don't have an easy way to communicate AB test buckets via an API. It may be the case that I just send the flags in this array, and determine the bucket from those flags in some enrichment step, or even just in the consumer. If we do this, it would make sense for flags to be an `Object` for our uses.

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+        "referrer": "...", 			// See also [5]
+		"type":	"content"			// Some sort of controlled list  
+    },
+	"meta": {						// [7]
+		"anything": "goes here",
+		"could: [ "be", "an", "array" ],
+		"or" : { "might": { "be": ["an", "object"] }}
+	},
+    "time": {
+         "offset": 234				// For offline stuff, I think
```
georgecrawford: Not just that, but any failed request, which will be queued by o-tracking and resent with an updated offset value.

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+{
+    "event": {
+        "source": "email.simple-email-service",  				// Name of the sender's system [1]
+        "type": "opened",  										// The type of event that happened, Eg. 'pageview', 'dwell', 'play' etc.
+		"api-key":	"0f7464b4-3f4d-11e4-984b-00144feabdc0"		// Sender-specific key [6]
+    },
+    "user": {													// Who generate the event
+        "user-agent": "Mozilla ...",
+        "ip": "123.123.123.123",
+        "anonymous-id": "...",									// An anonymous token, supplied by the client [2]
```
georgecrawford: @markstephens - we currently have a `deviceid` in ft-app, which is persisted in localStorage across user login/logout. Is this the same thing?

georgecrawford:
Other bits and pieces we currently send to iJento, which might be useful to codify:
- campaign/segId stuff - no idea if this is still useful
- we have an `appregion` value - might this be an FT-wide enum which could be useful?
- screen dimensions?
- we also have `requestid`, which allows page impressions and page events which resulted from a single user interaction to be grouped together.

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+- If `content.uuid` is included it will be expanded in to a set of content api meta-data.
+- ...
+
+## Format
+
+```
+{
+    "event": {
+        "source": "email.simple-email-service",  				// Name of the sender's system [1]
+        "type": "opened",  										// The type of event that happened, Eg. 'pageview', 'dwell', 'play' etc.
```
triblondon: Is there a taxonomy of these, or is this free text?

docs/event.md
```diff
@@ -0,0 +1,93 @@
...
+		"api-key":	"0f7464b4-3f4d-11e4-984b-00144feabdc0"		// Sender-specific key [6]
+    },
+    "user": {													// Who generate the event
```
triblondon: I would assume that in browser contexts the entire `user` object would be added by o-tracking

georgecrawford: The format certainly still needs documenting for the API, as we'll be sending from PHP for a while.

docs/event.md
georgecrawford: IP, of course, would have to be from a header.

docs/event.md
georgecrawford: I'm very interested in that answer. It's potentially a slippery slope and could add a delay to getting this format standardised, but may help enormously with interpreting the data the other end. @markstephens thoughts?

docs/event.md
commuterjoy: Yup, fair point. My goal here is to track actual rendered breakpoints and not move away from artificial divisions of mobile / tablet / desktop.

georgecrawford: Fine but we have our own non-standard rendered breakpoints. I suspect we should continue to send screen dimensions as well.

docs/event.md
commuterjoy: Yup. either way really. The only advantage of having it in the actual JSON as opposed to me grabbing it from the header would be to allow clients to specify/override it themselves, which I understand might be the case sometimes.

docs/event.md
commuterjoy: I think it can start off as list of conventions that you should abide by. I think if the consumers of the event stream want to filter out unknown events then that's up to them.

Also, the source gives us some context to help avoid conflicts - Eg, 'opening' the web app Vs. 'opening' an email

Essentially everyone wants a 'view' to mean the same thing everywhere, so I don't think there will be any argument in aligning the vocabulary.

docs/event.md
commuterjoy: Mostly, though Spoor will add the geography properties. And in Next-land the AB segment is pull from an API, and 'isStaff' is extracted from DAM on the server.

docs/event.md
commuterjoy: Yep, if for nothing else than to determine the breakpoints we might need to design for in the future.

docs/event.md
commuterjoy: ok, got it.

docs/event.md
commuterjoy: At the moment the CDN annotates each request with the GeoIP from MaxMind based on the IP of the sender.

In the future, perhaps :-

 - The above is a good default.
 - If the client specifies an IP address in the request header or JSON payload the enrichment process will have to do a MaxMind lookup.
 - If the client specifies the `geo` property in the JSON payload then no look-up needs to be done.

docs/event.md
georgecrawford: I suppose I mean that anything added during enrichment that ft-app will have to add before Spoor ingest for the time being should be documented, so that we conform and so there's no change when we move to o-tracking.

docs/event.md
commuterjoy: qatool and __toggler are the same thing I think (cookie based flags) so this makes sense.

>  it would make sense for flags to be an Object for our uses.

Agree, it should be the same as the AB test segments - i.e. `user.flags -> { "flag": "on", "foo": off }`

docs/event.md
georgecrawford: sounds good

docs/event.md
markstephens: For the most part, the browser will only send IDs for the user (session/guid/erights/passport/whatever)
This avoids any PII in the request and also means the product doesn't have to make user data available to o-tracking.

Also makes it easier for "roll-your-own" clients to integrate.

docs/event.md
markstephens: UA and IP will come from the headers.

As Matt says, there are case where they won't and need to be specified in the JSON manually.

docs/event.md
markstephens: Yes, I think that is what Matt means here.

docs/event.md
georgecrawford: Good to have confirmation, as our PHP will need to forward these headers :/

lucas42:
Could we have a field for indicating whether data is "live" or not?  It'd be great if we could send events from lower environments to the production version of spoor without affecting the real data used by the business.  This would allow our automated tests to look at the Kinesis stream and verify that our events are being properly ingested.

georgecrawford:
@lucas42 see line 82: "The general convention might be something like `product.sub-system.environment` - Eg, 'webapp.render.uat'"

commuterjoy:
@lucas42 Yes. I think you can do that by modifying the `event.sender` value ...

The general convention might be something like `product.sub-system.environment` - Eg, 'webapp.render.qa', 'webapp.render.prod'

lucas42:
So to filter out non prod-traffic, any consumer is going to have to parse the source value and work out what environment it's coming from?
Seems like unnecessary faff for each consumer to have to implement.

It also seems very tied to old notions of "environments".  What about services using blue/green deployments or other strategies which haven't been thought up yet?  When a service gets switched to live, it's going to have to update its 'source' to `foo.bar.prod`, even thought its architecture doesn't actually have 'prod'.

georgecrawford:
Do you have a suggestion?

lucas42:
I think there should be a separate field for indicating live traffic, rather than tying it to source.

I don't really mind what it's called or what the default should be if it's absent.
Possible options: `"islive": false` or `"testdata": true` or `"provenance": "live"`

markstephens:
I don't like all events going to the same endpoint and then being filtered later
- it seems prone to error
- as Luke mentioned, each consumer would have to implement a filter
- and with the quantity of data, it wouldn't be very efficient.

Having separate environments is also error prone, we've experienced the wrong stuff going to the wrong env in the past, but it immediately solves the filtering and efficiency issue.

markstephens:
In general this model is ok, I like that the structure is becoming more defined and things are having a place of their own, but it has some issues:

**Large issues**
* Type: Compared to o-tracking's latest model, we've lost category and action and replaced with type. Category and action did give a 2 level grouping for events, which is better for performance in Redshift (fewer queries having `LIKE 'something%'`) and also easier to group events ("show me all events in the same category")
* Low barrier to entry - A low barrier is fine, but it can't be so low the data is useless - suggest to make `type` mandatory too (or category & action if that sticks!) - as, by definition, an event is "doing" something, so type/category/action makes sense to include too.
* Hierarchy: Need a space to put ID's and parent ID's - suggest the `event` tag.

**Medium issues**
* UA and IP belong in device.
* Why are `orientation` and `layout` specifically mentioned? Surely there will be an event "changed orientation" and therefore just bung the values in the meta area?

**Small issues**
* anonymous-id - I had `spoor-id` - it's more explicit this ID will be set by the spoor eco-system (be that o-tracking or spoor-api.)
* Flip `event` and `meta` labelling - For me, `meta` should be info *about* the event. `event` should contain data *for* the event

**General comments**
* Predefined lists have been mentioned for some fields - like `type`. I'm relaxed about keeping it somewhat free. If we have something similar, a new product/thing should match, obviously, but to keep with the low barrier to entry and to allow someone to *just try something*, they should be allowed to send arbitrary values. As long as the json structure is intact, the new values will come out in both the stream and in redshift reporting without any effort.
* How do we decide which fields are *privileged* enough to become their own section in the JSON? For example, we will all agree the `content` section should be in there, but in reality it will appear in maybe 20% of all events. Category and action fulfils more use-cases for example.
* Enrichment: I thought the plan was to have all the enriched data in a separate area, often we want the original values as well, and don't want the risk of enrichment to overwrite anything.
* Anything which requires 2 teams to sync their release processes is bad, so things like:
    * mandated lists
    * anything needing a process (such as splitting source by dot to get an environment) - much better to be explicit.

commuterjoy:
> It also seems very tied to old notions of "environments"

True, but no matter how it works you are going to have to have something that represents code running in non-production that changes when real people are using it. I don't mind an explicit property called 'production'.

> Having separate environments is also error prone, we've experienced the wrong stuff
> going to the wrong env in the past, but it immediately solves the filtering and efficiency issue.

Yep, there's nothing stopping someone spinning up a test stack if they want to test something. That's safest.

There are no guarantees with anything we are doing that someone won't fill the production database full of junk.

commuterjoy:
> we've lost category and action

Yes. I like 'action' better.

> Categories

I don't like imposing a hierarchy on things. If someone want a to add a load of categories for their domain they can do that in meta.

>  suggest to make type mandatory too (or category & action if that sticks!) -

Agree.

> Hierarchy: Need a space to put ID's and parent ID's - suggest the event tag.

Meta, if we need them.

> UA and IP belong in device.

UA - I agree.

IP I guess is a property of the device's internet access point.

> Why are orientation and layout specifically mentioned? 

It's orientation and layout of the screen at the time the event was generated.

>  I had spoor-id

Agree. Though if some system has it's own anon id scheme I wanted to have some place to put that.

> Flip event and meta labelling

Yup, let me have a think about naming. I don't really like meta much. Maybe 'source' and 'event'?

> How do we decide which fields are privileged enough to become their own section in the JSON?

Here's my thinking :-

- The Source model describes generator (a computer system) of the event.
- The User model describes information about who the event was generated by.
- The Content model describes where the event happened, typically a webpage.
- The Time model describes when the event happened.
- The Event model allows additional context to be associated with an event.

> Enrichment: I thought the plan was to have all the enriched data in a separate area,

Could do, but I think it's simpler to merge it all in to one.

> Anything which requires 2 teams to sync their release processes is bad, so things like:

Agree.

georgecrawford:
Discussed just now with Mark - our `channel` field currently looks like `HTML5 r20150612.40259`. If we start using `event.source`, we'd also very much like to continue to sent `event.source.version = r20150612.40259`.

commuterjoy:
> HTML5 r20150612.40259

Yep.  Good idea. Next should probably submit build numbers too.

I don't like the idea of overloading `channel` etc. with more information, so splitting off to another property sounds ideal.

commuterjoy:
Had a think about the test data thing. I think a header `x-spoor-test-data: true` is the best option.

- Clients don't need to send data to different places or mutate their messages per environment.
- Responsibility to discard test data sits with the consumer.
- Nobody needs to maintain a separate infrastructure to test stuff.

lucas42:
`x-spoor-test-data` is fine with me.  (Though I don't see how varying a header is any different for the client than varying a field in the body)  I assume the spoor API is just going to take that header and put it in the body so the consumers can see it?

(Though if you want to be particularly pedantic, that naming convention for http headers has been deprecated: https://tools.ietf.org/html/rfc6648 )

markstephens:
Ultimately, this application and pipeline is going to sit in the Data Team's suite of applications at the end of all this. We are used to 2 environments so this will have 2 environments too.

Changing a header, or a value in the json is no different to changing a URL, and it saves every consumer of the queue getting confused and having to filter out all the test messages.

commuterjoy:
> getting confused and having to filter out all the test messages.

I see the test data header flag is just an explicit - 'throw this away' - flag for any consumer. Consumers will presumably have their own mechanisms to detect fraud, duplicates, outliers, bad data etc. I think an explicit test flag is just another factor in deciding on whether to take the data or chuck it away.

> We are used to 2 environments so this will have 2 environments too.

Yup, if someone wants to maintain two of everything that's entirely up to them :)

M

markstephens:
@commuterjoy @AlbertoElias 

I've written up the json structure after Matt and I discussed the other day:
https://github.com/Financial-Times/o-tracking/blob/flat-model/docs/event.md &
https://github.com/Financial-Times/o-tracking/blob/flat-model/docs/enrichment.md

This PR now has these docs, plus code changes to make o-tracking match this structure.

AlbertoElias:
No complaints on our side, looks good

PR #33, opened 2015-06-16
Structure

markstephens:
- Preparing for new structure.

### DIFF
--- a/main.js
+++ b/main.js
@@ -8,12 +8,24 @@ var settings = require("./src/javascript/core/settings");
  * @type {string}
  */
 var version = '0.0.24';
+/**
+ * The source of this event.
+ * @type {string}
+ */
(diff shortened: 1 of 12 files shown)


main.js
```diff
@@ -8,12 +8,24 @@ var settings = require("./src/javascript/core/settings");
...
+/**
+ * The source of this event.
+ * @type {string}
+ */
+var source = 'o-tracking';
+/**
+ * The API key.
+ * @type {string}
+ */
+var api_key = 'qUb9maKfKbtpRsdp0p2J7uWxRPGJEP';
```
AlbertoElias: is making an api key public ok?

markstephens: Yes, almost intentional!

It will allow us to rotate keys in the future and stop getting requests from some websites who like to copy our content.

PR #34, opened 2015-06-18
IE fixes with cross domain requests

markstephens:
@AlbertoElias I think I've fixed IE once and for all...!

### DIFF
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -21,7 +21,7 @@ var queue;
  */
 var currentRequests = {};
 
-function createXMLHttp() {
+function createTransport() {
 	try {
 		var xmlHttp = new window.XMLHttpRequest();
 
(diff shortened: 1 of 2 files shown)


src/javascript/core/send.js
```diff
@@ -29,29 +29,51 @@ function createXMLHttp() {
...
 		}
-
-		// Otherwise, check if XDomainRequest.
-		// XDomainRequest only exists in IE, and is IE's way of making CORS requests.
-		if (!utils.isUndefined(window.XDomainRequest)) {
-			return {
-				xmlHttp: new window.XDomainRequest(),
-				XDomainRequest: true
-			};
+	} catch (error) {}
```
AlbertoElias: Why do you suppress errors entirely?

markstephens: It doesn't matter if there is an error, it will fall back to the image tag method just below.

AlbertoElias: ok, feel free to merge then

PR #35, opened 2015-07-01
Docs

markstephens:
@AlbertoElias - they're not very detailed docs, but at least they're not wrong now...


README.md
```diff
@@ -1,170 +1,110 @@
...
 
-Origami module for the FT tracking.
+This module should be included on your product to make sending tracking requests to the Spoor API easy.
 
+## The Spoor ecosystem
 ![ScreenShot](https://rawgit.com/Financial-Times/o-tracking/master/resources/images/ngda-system-design.svg)
 
-**NOTE: This module is not yet functional. Please continue to use legacy tracking code (iJento in most cases)**
-
 # Usage
```
AlbertoElias: This should probably be ## as the title has the #

demos/json-config.html
```diff
@@ -14,5 +14,15 @@
 
 <script src="/bundles/js?modules=o-tracking:/demos/src/demo.js"></script>
 <script src="//registry.origami.ft.com/embedapi?autoload=resize"></script>
+
```
AlbertoElias: Regarding demos, you need to add the markup to a mustache file (if you're going to only build one demo for now, remove the other three), and the configure the config.json file for that demo, and use obt demo to build it

PR #38, opened 2015-07-08
Implemented component_id generation and adds tests

AlbertoElias:


src/javascript/event.js
```diff
@@ -49,14 +46,150 @@ function event(obj, callback) {
...
+ * @param {number} seed Positive integer only
+ *
+ * @return {number} 32-bit positive integer hash
  *
- * @param CustomEvent The CustomEvent
  * @private
  */
-function listener(e) {
-	event(e.detail);
+function _generateHash(str) {
```
samgiles: Must keep original license.

AlbertoElias: Arg! I removed that bit by mistake! Thanks!

PR #39, opened 2015-07-09
Docs

markstephens:


demos/src/json-config.mustache
```diff
@@ -0,0 +1,13 @@
...
+<script type="application/json" data-o-tracking-config>
+    {
+        "server": "https://test.spoor-api.ft.com/px.gif",
+        "context": {
+            "product": "origami" // e.g. webapp, next - This is a defined list, send the correct value!
+        }
+    }
+</script>
+
+<button onclick="var event = new CustomEvent('oTracking.page', { url: document.URL }); window.dispatchEvent(event);">Track a page</button>
```
AlbertoElias: could you make that document.body? I think o-tracking should attach to the body. And if it should still attach to window, it's better if a demo doesn't recommend products to dispatch it on window

demos/src/json-config.mustache
```diff
@@ -0,0 +1,13 @@
...
+<button onclick="var event = new CustomEvent('oTracking.event', { category: 'element', action: 'click' }); window.dispatchEvent(event);">Track an event</button>
```
AlbertoElias: Might be good to dispatch it from the button

markstephens:
@AlbertoElias - I've taken your suggestions for dispatching from body and the element in the demos.

PR #40, opened 2015-07-13
Fix root_id generation and correlation to match web app iJento calls

georgecrawford:


markstephens:
@georgecrawford Looks good to me!

PR #41, opened 2015-07-14
Events inherit `init` context

georgecrawford:


markstephens:
@georgecrawford This also looks good to me.

georgecrawford:
My only thought was whether `settings.get()` should always return a copy of the stored value, rather than a reference? Then we wouldn't need the complexity of `utils.merge({}, settings.get('config') && settings.get('config').context);`.

markstephens:
I don't mind, I guess it's easier to understand in the long term.

If we do that, we should be consistent and do the same with the other values in settings.

georgecrawford:
I think it's clearer this way. The `clone` implementation is a massive hack, but allows us to support basic scalars, native Array and native Object, which is probably enough.

PR #42, opened 2015-07-14
Remove unused counter param

georgecrawford:


markstephens:
@georgecrawford Yes, this is fine too.

PR #49, opened 2015-07-29
Capture device layout (o-grid) and orientation (o-viewport)

commuterjoy:
Next needs this (please).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,11 +1,17 @@
 {
   "name": "o-tracking",
-  "main": ["main.js"],
+  "main": [
+    "main.js"
+  ],
   "ignore": [
     "node_modules",
     "resources",
     "test",
     ".idea"
   ],
-  "description": "Origami module for FT tracking."
+  "description": "Origami module for FT tracking.",
+  "dependencies": {
+    "o-grid": "~3.1.8",
+    "o-viewport": "~1.4.0"
+  }
 }
--- a/src/javascript/core.js
+++ b/src/javascript/core.js
@@ -87,7 +87,9 @@ function track(config, callback) {
 		device: {
 			spoor_session: Session.session(),
 			spoor_id: User.userID(),
-			user_agent: window.navigator.userAgent
+			user_agent: window.navigator.userAgent,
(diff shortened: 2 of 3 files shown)


tom-parker:
:+1:

markstephens:
Isn't this a Next specific thing? e.g. WebApp aren't using grid or viewport.
If you already have grid and viewport on the page, you can pass these values through to o-tracking without it having to be a core part of the component:
```
otracking.init({
  ...
  device: {
    layout: require('o-grid').getCurrentLayout(),
    orientation: require('o-viewport').getOrientation()
  }
  ...
});
```

I agree we should have a standard implementation for dimensions though, maybe this commit is it, I just don't know if these are standard, or bloat the component out or anything else.

kaelig:
Note that `getCurrentLayout` relies on values surfaced to JS via CSS. It works only if a specific part of o-grid's CSS is output.

georgecrawford:
I'm with Mark on this. I'd like to avoid bloating o-tracking as much as possible. However, I can see the value of having viewport information tracked by default so that other consumers can drop in o-tracking and have it all sent automatically.

Is there value in an optional extension package/mode for o-tracking, which would send a larger bundle of JS (potentially including o-viewport, etc.), and report all the useful environmental characteristics we might need? More complex products such as the web app could then continue to pass these in manually using the lightweight base library.

markstephens:
Then, for me, these shouldn't be a core part of this component.
We could make them a standard for tracking dimensions though.

commuterjoy:
ok, I agree. 

https://github.com/Financial-Times/next-js-setup/pull/67

markstephens:
Shall we add to the docs that if you're sending layout and orientation, these should come from the origami components so that we're standardised?

JakeChampion:
I thought it already was?
https://github.com/Financial-Times/o-tracking/commit/bd7073ab23418ef8c00beb95b0e9d0be956a58cb#diff-e10b07bee0888ffd952c5c784e7a87e8R25

georgecrawford:
It'd be good to specify `orientation` as an enum of `portrait|landscape`, as I think it's safe to standardise on those.

PR #50, opened 2015-08-03
Document is-offline flag

georgecrawford:
In ft-app, we make a best-guess as to whether the User Agent is offline or not, and vary some behaviour accordingly. It's useful to us to document this state along with each event.


docs/event.md
```diff
@@ -23,11 +23,12 @@ This allows a low barrier to entry for anyone wanting to submit events to Spoor.
...
 		"ip": "123.123.123.123",
 		"layout": "S",											// require('o-grid').getCurrentLayout
-		"orientation": "portrait"								// require('o-viewport').getOrientation
+		"orientation": "portrait",								// require('o-viewport').getOrientation
 		"dimensions": {
 			"width": 1024,
 			"height": 768
-		}
+		},
+		"is-offline": true
```
triblondon: Use of hyphen seems inconsistent with spoor_id, ft_session, user_agent etc.

georgecrawford: Thanks, totally got that wrong!

commuterjoy:
Can this be inferred by the presence of a the time `offset`?

georgecrawford:
I don't think they're quite the same thing, although they will often correlate. `offset` can simply indicate a failed network request, and even if that is because the device is offline, we may not have correctly determined that fact. We try to derive the `is-offline` state, and vary behaviour on that, so it's quite important for us to track it.

markstephens:
I much prefer to have a specific value for things than have many people trying to recalculate something, so it's a :+1: from me.

PR #56, opened 2015-08-14
Cookie fix

markstephens:
Clean up after forgetting to write cookies to the 'root' ft.com domain.
- Check local storage for the 'proper' value.
- If it exists, use it.
- If not, set current user id as the 'proper' value.
- If this value and the cookie match, then we've already fixed it.
- If not, drop the cookie and it will be reset it on the root domain.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,7 +1,7 @@
 /*global require, module */
-"use strict";
+'use strict';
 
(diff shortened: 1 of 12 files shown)


markstephens:
:+1: Tested functionality in real-world situation.

Issue #93, opened 2015-12-04
Merge n-instrumentation in to o-tracking


wheresrhys:
- [x] o-grid/o-viewport - decorate all events with current layout and orientation https://github.com/Financial-Times/n-third-party-code/blob/master/src/analytics/o-tracking-wrapper.js
- [x] copy - to be migrated unchanged. Look into Tynt's functionality to make sure the events o-tracking sends could, in theory, be generated by tynt if we need to swap it in in future
- [x] scroll depth - track 10, 20 etc percentiles by default. Add ability to declaratively specify an element is interesting to track scroll for using attributes `data-o-tracking--scroll-bottom`, `data-o-tracking--scroll-top`. Use o-viewport to throttle scroll events
- [x] nav timing - bundle with page view event if possible
- [x] attention time - port implementation from https://github.com/Financial-Times/next-js-setup/commit/ee489d92c99d0c9ba864ca0ac38d76b1f3b1b853
- [x] domPath - propose a middle ground between next's terse, human readable domPaths and origami's hashed, xpath based element id's for identifying where on the page an element is. 
- [x] Try to unify link and cta tracking
- [x] mouse tracking - Consider tracking physical location of clicks https://github.com/Financial-Times/n-instrumentation/pull/17
- [x] use o-viewport for visibility https://github.com/Financial-Times/o-viewport/issues/25

@markstephens @alicebartlett @triblondon @commuterjoy

commuterjoy:
> Look into Tynt's functionality to make sure the events
> o-tracking sends could, in theory, be generated by tynt if we need to swap it in in future

on next I don't think we want Tynt directly to be manipulating URLs etc. or sending events.

> human readable domPaths

please, please keep something usable and understandable to 99% of the population :)

wheresrhys:
Dependent on
- https://github.com/Financial-Times/o-grid
- https://github.com/Financial-Times/polyfill-service/pull/536
- https://github.com/Financial-Times/o-viewport/issues/25 (sort of)

wheresrhys:
So here's my domPath/element id proposal

## Config
- includeAllNodes (Bool) - whether to include every dom node in the element genertaed id (if false only includes nodes with data-trackable attributes) [default false]

If the above is true then id's will be the result of traversing the DOM, including sibling index. Not recommended at all!

If the above is false the following attributes can be put in the DOM

## Attributes
[...]

PR #94, opened 2015-12-07
Changing directory structure

wheresrhys:
The idea being that it provides a good foundation to gradually start importing the more diverse range of events available in n-instrumentation (https://github.com/Financial-Times/o-tracking/issues/93 is turning into a bit of a monster)


markstephens:
Hi @wheresrhys 

Thanks for the changes, it looks good to me :+1:

Mark

PR #96, opened 2015-12-17
added send beacon transport

wheresrhys:
As part of this I've stopped passing anything in to the on complete callback. It's not documented what gets passed into this callback by o-tracking (and it varies depending on whether the transport is image or XHR anyway) so I think it's, if not necessarily safe, technically fine to stop passing anything in. Better might be to return a Promise, resolved or rejected on request completion/error. Depending on how careful we want to be around people's use of undocumented parameters passed to callback this may necessitate a major release, but would be a step forward worth making I think.

### DIFF
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -28,7 +28,7 @@ let queue;
 function sendRequest(request, callback) {
 	const offlineLag = (new Date()).getTime() - request.queueTime;
 	let path;
-	const transport = navigator.sendBeacon ? sendBeacon() : window.XMLHttpRequest && 'withCredentials' in new window.XMLHttpRequest() ? xhr() : image();
+	const transport = (navigator.sendBeacon && Promise) ? sendBeacon() : window.XMLHttpRequest && 'withCredentials' in new window.XMLHttpRequest() ? xhr() : image();
 	const user_callback = request.callback;
 
 	const core_system = settings.get('config') && settings.get('config').system || {};
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -91,9 +91,11 @@ function sendRequest(request, callback) {
  */
 function add(request) {
 	request.queueTime = (new Date()).getTime();
-
-	queue.add(request).save();
-
+	if (navigator.sendBeacon && Promise) {
+		sendRequest(request);
+	} else {
+		queue.add(request).save();
+	}
 	utils.log('AddedToQueue', queue);
 }
 
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -30,7 +30,7 @@ function sendRequest(request, callback) {
 	const queueTime = request.queueTime;
 	const offlineLag = new Date().getTime() - queueTime;
 	let path;
-	const transport = (navigator.sendBeacon && Promise) ? transports.get('sendBeacon')() :
+	const transport = (navigator.sendBeacon && Promise && settings.get('useSendBeacon')) ? transports.get('sendBeacon')() :
 										window.XMLHttpRequest && 'withCredentials' in new window.XMLHttpRequest() ? transports.get('xhr')() :
 										transports.get('image')();
 	const user_callback = request.callback;
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -3,7 +3,9 @@
 const settings = require('./settings');
 const utils = require('../utils');
 const Queue = require('./queue');
-
+const xhr = require('./transports/xhr');
+const sendBeacon = require('./transports/send-beacon');
+const image = require('./transports/image');
 /**
  * Default collection server.
  */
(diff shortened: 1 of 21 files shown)


src/javascript/core/send.js
```diff
@@ -24,66 +26,7 @@ let queue;
...
-							callback(null, xmlHttp);
-						} else {
-							callback('Incorrect response: ' + xmlHttp.status, xmlHttp);
-						}
-					};
-				}
-			};
-		}
-	} catch (error) {
-		utils.broadcast('oErrors', 'log', {
```
wheresrhys: I don't know if this is all that useful.

src/javascript/core/transports/send-beacon.js
```diff
@@ -0,0 +1,24 @@
...
+        send: function (domain, path) {
+            if (navigator.sendBeacon(path, domain)) {
+                resolver();
+            } else {
+                rejecter(new Error('Failed to send beacon event: ' + domain.toString()));
+            }
+
+        },
+        complete: function (callback) {
+            console.log('asdjasghjg')
```
markstephens: remove console.log.

markstephens:
Thanks for the PR @wheresrhys 

I have 3 main issues & questions with this:
- the send beacon method uses promises.
    - you only check for the [presence of sendbeacon](https://github.com/Financial-Times/o-tracking/pull/96/files#diff-6542a5038b04b0f87f67d46eb8b3a6bdR31), but don't check promises functionality is ok for the browser
    - If promises don't work in the browser, a lot of the events may fail without us knowing why - I know we have the polyfil service, but it might be missed or whatever, and then we lose data.
- does send beacon send the message with the correct content type? spoor-api only accepts application/json for POSTs
- what's the advantage of doing all this change?

wheresrhys:
- good point, I'll add a check for promises (though I have a feeling all browsers that support sendBeacon also support Promises)
- I'll test out this branch on next before merging and releasing. I would expect the data type to be ok. The spec is quite opaque, but I believe it attempts to convert to JSON before falling back to string https://fetch.spec.whatwg.org/#bodyinit, and I'd imagine that would entail setting a content-type header
- sendBeacon already has a lazy sending & queuing mechanism built in. Once it's available for X% of our users then much of the queuing implementation built in to this component can be stripped out. I would also expect sendBeacon to be better at e.g. sending tracking events for users who are offline when they click a button, but don't subsequently revisit the site.

wheresrhys:
in fact I should refactor to not use the queu when sendBeacon is the transport, otherwise we lose the advantages it brings

markstephens:
> in fact I should refactor to not use the queu when sendBeacon is the transport, otherwise we lose the advantages it brings

Sometimes queue use is forced, when tracking links for example, as the event needs to know about the "source" and "destination" page

wheresrhys:
is that the only instance of that sort of behaviour? If so, we could force XHR+queue fro those only. Longer term it'd be nice to think of a way to create the link between source and destination which would be compatible with sendBeacon

markstephens:
Yes, it is the only instance of this.

wheresrhys:
It looks like link tracking has its own queue so (pending testing it out to verify it works) I think we're safe.

Remind me why links are tracked this way, as opposed to having a pair of events - the click and the subsequent page view. If spoor + new ngda tooling enabled querying the data to generate equivalent reports would it be feasible to drop this feature (not immediately, but schedule its demise)?

markstephens:
yes.

There are actually 3 events here, each page view is an event and the link event joins them together.
```
page 1              page 2
   |------ link ------|
```

AlbertoElias:
We should count on people using the polyfill service, which supports Promises. You do need to add it to your origami json though

markstephens:
> We should count on people using the polyfill service, which supports Promises. 

I'm starting to think we should take a different approach with tracking and code it for the worst browser possible.

It is getting frustrating to keep having to weed out different bugs in browsers. We are comparing o-tracking to our current analytics provider, and o-tracking keeps having bugs.
For example, at the moment, o-tracking isn't working for 12% of Safari 9 on iOS, but is working for the other 88% of safari 9 on iOS - this is a new browser (there are other browsers to sort out too, IE is pretty bad). The reason we know this, is because the current analytics provider is working in these browsers. Whilst it is old code, it is coded for the IE 6's of the world and seems to work in all the newer browsers too without modification.

I'm thinking introducing modern functionality for sake of it doesn't seem to be working. Examples:

- CTM CSS image fallback just doesn't work
    - Especially in IE (which is it's main usecase) it's double posting with the JS.
    - I've ended up doing it in JS to get around it e.g. `if (ctm) { otracking.init() } else { new Image().src = 'blah' }`
- Switching to ecma6 hasn't been too helpful
    - Maybe obt can run better linting and maybe it lets me write more readable code, but
    - disadvantages seem to be (the possibility of) a larger file size
    - And worse is that people consuming o-tracking have had to modify their build and introduce babel - a tracking library shouldn't be causing that much work.
- And now Rhys has spent a good few days introducing sendBeacon
    - which to me doesn't provide much benefit over what was in there and already working (maybe we can strip 20 lines out eventually)
    - but will introduce more uncertainty as to whether we're capturing everything, as it's using more unsupported or unfinished features.
- Not everyone will use the polyfil service, third parties might resist for example.

All of the above makes me wonder if we should code tracking for the basic browser. Soon we will lose the ability to compare and find our bugs, as we turn the current provider off, and then we won't know what data we're missing. All of it _should_ work, but in practice the results are telling me otherwise.

wheresrhys:
All good points. @alicebartlett if Mark's available can we discuss this at origami meeting tomorrow.

My general thoughts on reliability/bugs is it isn't a coding style problem but a QA one - as o-tracking is so critical it should have tests running in lots of browsers using saucelabs, proper test cases written, e2e as well as unit tests and maybe test coverage reporting (test coverage is currently somewhere between ok and good, but should be excellent). Could also introduce some kind of canary testing (a tiny portion of page views receive both the stable and next versions of oTracking to compare the figures)

And following on from that point, it's a resourcing problem. I imagine when iJento was in its heyday they would have had a lot more than 1-ish developer working on the client library part time.

If we get all the above right then we needn't worry so much about using new APIs, working well across browsers etc. The points about babel, polyfill service and build complexity... as with everything to do with delivering origami components to the browser I have mixed feelings.

With regard to sendBeacon the benefit is it's an API designed specifically for tracking and will be more reliable at tracking events when offline, or just before navigating away from the site. Could make xhr the default for now though, overridable in config. (I've been thinking the tests should use a mockTransport object anyway to decouple them from the actual send event implementation... would make cross browser test running easier)

alicebartlett:
@wheresrhys Yep this sounds like a good thing to discuss. I'll add it to the agenda if @markstephens can make the meeting (2pm)

markstephens:
Thanks @wheresrhys and @alicebartlett 

Good points Rhys, I agree with you. I appreciate the work you're putting into o-tracking.
Alice, I can't make the meeting tomorrow, I'm in 2 meetings at that time already!

alicebartlett:
Hello, OK, summarising that meeting we had here:

- in 1 month notice on iJento will be given, or the contract will have to be extended for another 3 months. It seems likely that we'll be able to terminate iJento and switch but there are some bugs in o-tracking that need to be fixed.
- we currently have two sources for tracking data, iJento and o-tracking, meaning that if a bug is introduced to o-tracking we can see it because we can see a drop compared to the numbers in iJento. Removing iJento leaves us vulnerable to introducing a bug in o-tracking (that, say only affects 3% of users) and then not noticing.
- o-tracking needs some time spent on firming up its test suit to mitigate this risk. We should also add browser testing, that runs the tracking in multiple clients. We could use sauce labs for this, which is what the polyfill service uses, or webdriver, which is what @georgecrawford is looking at using for the web app.

*actions*
- Once this pr is passing CI we can merge it, it makes more sense to merge it before iJento is switched off so it's possible to see if it works. @wheresrhys to get it to a mergable state.
- @wheresrhys and Akalan (a new developer to the data team) to work together on improvements to test coverage / browser coverage for o-tracking ahead of switching off iJento

PR #97, opened 2016-01-04
cookies on sub-domain

markstephens:
Punt on a fix for the issue where we're still seeing spoor-id cookies set against subdomains.

- Don't think cookies are this strict, but worth seeing if dropping the cookie with and without a path fixes the issue.


markstephens:
cancelled

Issue #81, opened 2015-10-28
Causes Fatal Error on windows phone 8 (IE11)


aintgoin2goa:
For some reason that I don't fully understand the way in which data is persisted in local storage seems to fail in IE11 on windows phone (and possible desktop - haven't checked)

When the queue is created it is saved as a HTML Encoded representation of an empty array ("%5B%5D" rather than "[]").  Therefore, this code (store.js line 138)
```
if (loadStore) {
		try {
			this.data = JSON.parse(loadStore);
		} catch (error) {
			utils.broadcast('oErrors', 'log', {
				error: error,
				module: 'o-tracking'
			});
			this.data = loadStore;
		}
	}
```
causes the queue to be set to "[]" rather than [].

This means that in this bit (queue.js, line 100):
```
function doAdd(item) {
		self.queue.push({
			created_at: (new Date()).valueOf(),
			id: utils.guid(),
			item: item
		});
	}
```
self.queue is a string not an array.  Therefore the whole operation fails.

adambraimbridge:
http://i.imgur.com/lBzg3NF.gif

markstephens:
This should be fixed with version 1.0.17

PR #101, opened 2016-01-06
Cookie storage bug

markstephens:
- Forgot to decode values in cookies, they're set as encoded strings
- Likely a fix for https://github.com/Financial-Times/o-tracking/issues/81 too


markstephens:
Tested this fix on ios and it's good.

PR #102, opened 2016-01-07
Fixes an issue in IE8 where the clone() is called with a value of und…

Thurst-ft:
@markstephens:  This seems to fixe the IE8 issue I mentioned on slack yesterday


markstephens:
:+1: 

Thanks @Thurst-ft

PR #108, opened 2016-01-28
only init listeners on init, not on load

wheresrhys:
The current implementation - where listeners are added on load - means that we can't turn o-tracking fully off with a flag as it still tries to send requests to spoor when something on the page fires an oTracking event. This throws an error because o-tracking hasn't been initialised and configured

```
Uncaught TypeError: Cannot read property 'read' of undefinedgetSession @ session.js:43track @ core.js:97event @ custom.js:67module.exports @ dispatchCustomEvent.js:16(anonymous function) @ navigation-timing.js:98setTimeout (async)load @ navigation-timing.js:21
```

@markstephens

### DIFF
--- a/main.js
+++ b/main.js
@@ -165,7 +165,8 @@ Tracking.prototype.init = function(config) {
 			}
 		}});
 	}
-
+	this.event.init();
+	this.page.init();
 	this.initialised = true;
 	return this;
 };
--- a/src/javascript/events/custom.js
+++ b/src/javascript/events/custom.js
@@ -218,6 +218,9 @@ function _generateHash(str) {
 	return h >>> 0;
 }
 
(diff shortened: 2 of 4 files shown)


markstephens:
:+1: but...

What happens to the events triggered before init? This is why google analytics has that array of events thing - [easiest to see here](https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet) - first line of code.

wheresrhys:
um, well, that'd be something to fix for sure, but at the moment it's broken anyway. 

I suppose on load o-tracking would create a queue to hold the raw events, undecorated with config, and then flush it on init. I'll raise a separate issue

wheresrhys:
https://github.com/Financial-Times/o-tracking/issues/109

PR #112, opened 2016-02-05
avoid InvalidStateErrors in IE10

wheresrhys:
see http://stackoverflow.com/a/19667959/87739. Hopefully this'll have a big effect in IE @markstephens


wheresrhys:
windows phone 8 reliably sends single page:view event with this

markstephens:
As simple as that! Blimey!

Thank you very much @wheresrhys

PR #116, opened 2016-02-15
Exception reporting

markstephens:
- IE10 can't handle putting an exception in a custom event.


wheresrhys:
I wonder if it'd be better to broadcast the error itself and not just the message. Not sure if o-errors would handle it, but I expect it would @AlbertoElias

PR #120, opened 2016-02-26
Update docs to reference v2 endpoint of the build service

onishiweb:


AlbertoElias:
All those use strict you've added makes linting fail, as use strict isn't necessary in a es6 module

onishiweb:
@AlbertoElias yeh it turns out I have a different version of ESLint than OBT uses, so I was getting an error for the missing `use strict`'s. Will fix it this morning.

PR #124, opened 2016-04-06
prevented custom event context being sent as context.context

wheresrhys:
To date if a custom event 

```
{
category: 'cat',
action: 'act',
context: {
  cont: 'cont'
}
```

is fired, it is sent as

```
{
category: 'cat',
action: 'act',
context: {
  context: {
    cont: 'cont'
  }
}
```

I've put some stuff in to continue to do that (for backwards compatibility), while also, more sensibly, sending the supplied context object as the context. To be honest, I think the current implementation, where the entire event detail is deliberately coerced into being the context, should be dropped if we're confident it won't mess up anything anyone is actively tracking.

@markstephens @Lucas-Gluchowski


markstephens:
Ah, I wondered why that was happening. It shouldn't double the context.


> To be honest, I think the current implementation, where the entire event detail is deliberately coerced into being the context, should be dropped if we're confident it won't mess up anything anyone is actively tracking.

Why? It's effectively just a namespace. The idea is to provide context about the event.

wheresrhys:
because if the docs, examples and other event types require a user to set a `.context: {}` property in order to send context, custom events should not be any different, otherwise it's too magical and confusing

markstephens:
Ok, in `init` there's a reference to context.

All these values are already so different between platforms, we need to sort that out first. At the moment it's rubbish.

wheresrhys:
> sort that out first

`context.context` feels like a definite bug we should fix ASAP before too many people start building queries based on it

wheresrhys:
I still think this is worth doing, but I don't think there's complete consensus on how/if context should be inherited from global settings, so pointless fixing the branch before that.

notlee:
Hi @wheresrhys, I'm not familiar with this or sure who this effects. Is this something that should be rekindled?

notlee:
I'm closing this PR as it's been hanging around a few years. Please let us know if it can be reopened and released.

PR #132, opened 2016-07-15
Link tracking

markstephens:
Match spec @ https://drive.google.com/open?id=18G4oEkfdMbG719YmMnIHeOcD8hVHLk4q9Dr-7pazxRk

### DIFF
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -22,7 +22,7 @@ let queue;
 /**
  * Consistent check to see if we should use sendBeacon or not.
  *
- * @return {boolean}
+ * @return {boolean} Should we use sendBeacon?
  */
 function should_use_sendBeacon() {
 	return (navigator.sendBeacon && Promise && (settings.get('config') || {}).useSendBeacon);
--- a/src/javascript/core/transports/image.js
+++ b/src/javascript/core/transports/image.js
@@ -17,4 +17,4 @@ module.exports = function () {
 			}
 		}
 	};
-}
+};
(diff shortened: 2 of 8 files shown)


georgecrawford:
Position and siblings?

markstephens:
Yeah, haven't got there yet, and Not sure I can work them out generically, probably the product has to tell me. They can be passed in already..

commuterjoy:
Next uses camel case FWIW.

Is this attached to all events on the page? Or should it / can it be initialised explicitly.

commuterjoy:
> and not sure I can work them out generically,

I don't think you can, but if the products decorates the DOM with position/sibling attributes then `o-tracking` can capture the values, Eg.

https://github.com/Financial-Times/n-ui/blob/master/tracking/ft/events/cta.js#L58

markstephens:
>Is this attached to all events on the page? Or should it / can it be initialised explicitly.

You can tell it to listen on all A tags (other elements too), or call it explicitly yourself (e.g. custom event or call the function directly)

> if the products decorates the DOM with position/sibling attributes then o-tracking can capture the values

Yes, that's true, but what about links that we haven't set those attributes for?

commuterjoy:
> Yes, that's true, but what about links that we haven't set those attributes for?

They just don't get captured. It's only an optional field.

markstephens:
Yes, ok, same as `data-trackable`

commuterjoy:
> Yes, ok, same as data-trackable

yup!

PR #152, opened 2017-03-01
Only allow callbacks that are functions.

adambraimbridge:
Something seems to be somehow calling o-tracking/send.js `run(callback)` with a `callback` that is an `Event` of `type:"online"`. 

image

I reproduced this in the `next-article` app by going offline (switching off wifi), waiting a minute, then coming back online.

I tried to find what was setting `callback` to an `Event`, with no success. So the best I can do for now is guard against any non-function callbacks.

This will hopefully reduce the "t is not a function" errors in [sentry](https://sentry.io/nextftcom/frontend/?query=is%3Aunresolved+t+is+not+a+function).

image

### DIFF
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -122,6 +122,9 @@ function add(request) {
 function run(callback) {
 	if (utils.isUndefined(callback)) {
 		callback = function () {};
+	} else if (typeof(callback) !== 'function') {
+		utils.log(`o-tracking error: 'callback' is defined, but is not a function.`)
+		callback = function () {};
 	}
 
     // Investigate queue lengths bug
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -122,9 +122,6 @@ function add(request) {
 function run(callback) {
 	if (utils.isUndefined(callback)) {
 		callback = function () {};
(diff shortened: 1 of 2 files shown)


src/javascript/core/send.js
```diff
@@ -122,6 +122,9 @@ function add(request) {
 function run(callback) {
 	if (utils.isUndefined(callback)) {
 		callback = function () {};
+	} else if (typeof(callback) !== 'function') {
+		utils.log(`o-tracking error: 'callback' is defined, but is not a function.`)
```
lc512k: Thanks Adam!!
Will these be logged beyond heroku logs? Sentry?
Also, would it help to add the `callback` itself to the logs for debugging?

adambraimbridge:
Supercalifragilisticextramegathanks to @JakeChampion (who by the way gave me a vocal 👍)

`utils.addEvent()` is a wrapper for `addEventListener()`. Event listeners pass an `Event` object as a parameter to their handler function. So this line: 

```javascript
utils.addEvent(window, 'online', run);
```

Made the window listen for any `online` event and call `run({event})` when triggered. `{event}` of course being an object — thus breaking `run(callback)` which expects `callback` to be a function.

PR #153, opened 2017-03-06
Securedrop stop tracking

StuartFT:
This change is added to support securedrop, after this change any clicks to the securedrop service should be excluded from tracking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "main": "main.js",
   "private": true,
   "scripts": {
-    "test": "./node_modules/karma/bin/karma start karma.conf.js"
+    "test": "./node_modules/karma/bin/karma start karma.conf.js",
+    "test-watch": "node node_modules/karma/bin/karma start --browsers=Chrome --single-run=false --auto-watch"
   },
   "devDependencies": {
     "babelify": "^6.1.3",
--- a/src/javascript/events/click.js
+++ b/src/javascript/events/click.js
@@ -121,9 +121,8 @@ const getEventProperties = event => {
 
 // Controller for handling click events
 const handleClickEvent = eventData => (clickEvent, clickElement) => {
-	//we don't want to track clicks to the securedrop anonymous service
-	if (typeof clickElement.href !== "undefined" &&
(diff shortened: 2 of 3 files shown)


commuterjoy:
Did we not discuss adding a more generic attribute.

adambraimbridge:
@commuterjoy: Something like `data-trackable-false` perhaps?

StuartFT:
@adambraimbridge 

pushed a change to use the data  attribute, just seen your comment, tried do you think below reads better? I'm thinking inline might be just a bit easier to read?

@adambraimbridge @commuterjoy data attribute name ok as `data-o-tracking-do-not-track` ?

Thanks, Stu

https://refactoring.com/catalog/extractMethod.html style:
```
const isOkToTrack = clickElement => {
	//we don't want to track clicks to anonymous services like securedrop
	return clickElement.getAttribute("data-o-tracking-do-not-track") === "true" ? true : false;
};

// Controller for handling click events
const handleClickEvent = eventData => (clickEvent, clickElement) => {
	if (!isOkToTrack(clickElement)) return;
```

or inline style (https://refactoring.com/catalog/inlineMethod.html) :)
```
const handleClickEvent = eventData => (clickEvent, clickElement) => {
        //we don't want to track clicks to anonymous services like securedrop
        if (clickElement.getAttribute("data-o-tracking-do-not-track") === "true") return;
```

adambraimbridge:
@StuartFT: In your commit, I agree inline's better than extracting it to a method. (Nice link BTW, interesting site).

However I was thinking more along the line of walking the whole DOM from the clicked element through its parents up to the HTML node. Like how [getTrace](https://github.com/Financial-Times/o-tracking/blob/master/src/javascript/events/click.js#L79) does it. 

In that way, a `data-o-tracking-do-not-track ` attribute doesn't have to be applied to every non-trackable element explicitly, but can be added to any of its parents. I was overthinking it, to be honest.

Your PR is simple and works. 👍 from me. 

(I've no opinion on the name of the `data-o-tracking-do-not-track` attribute, so long as it's used consistently.)

PR #157, opened 2017-03-14
Allow tracking settings to be changed after init to allow for changes during the lifetime of a page, merge in device config to tracking calls, and rework core track logic

rowanbeentje:
These are the base changes required for the app and its single-page model to work with o-tracking; allowing the configuration to be updated without recreating the object means that user etc state can change and be reflected in tracking calls.

Adding context, device, and user from the base config consistently to tracking calls simplifies the code, and means device details can easily be supplied for the first time (we assume this is why Next is currently shoehorning device properties into the user object?).

More than I intended to do but far less than I wanted to do after I started!  Will annotate some parts of the PR...

### DIFF
--- a/main.js
+++ b/main.js
@@ -115,33 +115,21 @@ Tracking.prototype.init = function(config) {
 
 	const hasDeclarativeConfig = !!this._getDeclarativeConfigElement();
 
-	if (!(hasDeclarativeConfig || config)) {
-		return this;
-	}
-
 	config = config || {};
 	if (hasDeclarativeConfig) {
 		config = this._getDeclarativeConfig(config);
 	}
 
-	settings.set('config', config);
 	settings.set('version', this.version);
 	settings.set('source', this.source);
 	settings.set('api_key', this.api_key);
 
 	settings.set('page_sent', false);
 
-	// Developer mode
-	if (config.developer) {
-		this.developer(config.developer);
+	// Set up the user from stored - may later be updated by config
+	user.init();
 
-		if (config.noSend) {
-			settings.set('no_send', true);
-		}
-	}
-
-	// User identifier
-	user.init(config.user ? config.user.user_id : null);
+	this.updateConfig(config);
 
 	// Session
 	session.init(config.session);
@@ -153,6 +141,34 @@ Tracking.prototype.init = function(config) {
 	this.page.init();
 	this.initialised = true;
 	return this;
+};
+
+/**
+ * Update the tracking configuration with any state changes. The supplied
+ * config is merged with any existing config; to unset a value, set it as
+ * null or undefined.
+ *
+ * @param {Object} newConfig The configuration object to merge in - see init()
+ * @return {void}
+ */
+Tracking.prototype.updateConfig = function(newConfig) {
+	let config = settings.get('config') || {};
+
+	config = this.utils.merge(config, newConfig);
+	settings.set('config', config);
+
+	// Developer mode
+	if (config.developer) {
+		this.developer(config.developer);
+
+		if (config.noSend) {
+			settings.set('no_send', true);
+		}
+	}
+
+	if (config.user && config.user.user_id) {
+		user.setUser(config.user.user_id);
+	}
 };
 
 /**
--- a/src/javascript/core.js
+++ b/src/javascript/core.js
@@ -6,69 +6,31 @@
 const Send = require('./core/send');
 const User = require('./core/user');
 const Session = require('./core/session');
-
-/**
- * Shared 'internal' scope.
- * @type {Object}
- */
 const settings = require('./core/settings');
 const utils = require('./utils');
 
-/**
- * Default properties for sending a tracking request.
- * @type {Object}
- * @return {Object} - The default settings for the component.
- */
-const defaultConfig = function () {
-	return {
-		async: true,
-		callback: function () {},
-		system: {},
-		context: {},
-		user: {
-			passport_id: utils.getValueFromCookie(/USERID=([0-9]+):/) || utils.getValueFromCookie(/PID=([0-9]+)\_/),
-			ft_session: utils.getValueFromCookie(/FTSession=([^;]+)/)
-		}
-	};
-};
+let rootID;
 
 /**
- * Generate and store a new rootID.
- * @param {string} new_id - Optional rootID, if you want to use your own. Otherwise we'll create one for you.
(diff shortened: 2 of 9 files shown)


main.js
```diff
@@ -115,33 +115,21 @@ Tracking.prototype.init = function(config) {
 
 	const hasDeclarativeConfig = !!this._getDeclarativeConfigElement();
 
-	if (!(hasDeclarativeConfig || config)) {
```
rowanbeentje: I moved this to far lower - as far as I can tell, if no config is supplied there's no reason not so still set up the class correctly with version and required helpers.

src/javascript/core.js
```diff
@@ -6,69 +6,31 @@
...
- */
 const settings = require('./core/settings');
 const utils = require('./utils');
 
-/**
- * Default properties for sending a tracking request.
- * @type {Object}
- * @return {Object} - The default settings for the component.
- */
-const defaultConfig = function () {
```
rowanbeentje: I significantly refactored this core tracking class to simplify the logic.  I don't see an advantage of having some default config up here and some down in the `track` function.

src/javascript/core.js
```diff
@@ -6,69 +6,31 @@
...
- * Generate and store a new rootID.
- * @param {string} new_id - Optional rootID, if you want to use your own. Otherwise we'll create one for you.
+ * Generate and store a new rootID, used to mark a new root event that
+ * subsequent events will be linked to.
  * @return {string|*} The rootID.
  */
-function setRootID(new_id) {
-	settings.set('root_id', requestID(new_id));
-	return settings.get('root_id');
+function setRootID() {
```
rowanbeentje: While this is technically available externally if something `require()`s in the core object separately, it's not documented and I think it shouldn't be - this is internal behaviour and there's no advantage to allowing it to be customised, and just adds complexity.  It can still be overridden by supplying it as part of the event config.

src/javascript/core.js
```diff
@@ -83,35 +45,48 @@ function track(config, callback) {
 	if (utils.isUndefined(callback)) {
 		callback = function () {};
 	}
-
-	const coreContext = settings.get('config') && settings.get('config').context || {};
```
rowanbeentje: I had troubles tracing out the various directional merges happening across settings context/config context/default config at the top/parameters.  I hope the new layout makes order of overrides clearer.

src/javascript/utils.js
```diff
@@ -49,7 +49,7 @@ function is(variable, type) {
 }
 
 /**
- * Merge objects together. Will remove 'falsy' values.
+ * Merge objects together. Will remove undefined and null values.
```
rowanbeentje: I've updated this to reflect what the code is actually checking for.  But this isn't quite right either - what it's really doing is ignoring undefined and null values in the object being merged in.  So there's no way to delete an existing key right now, while from this I assumed that using a falsey value would unset the key from the merged object...

test/core.test.js
```diff
@@ -62,11 +71,11 @@ describe('Core', function () {
 			// Context
 			assert.deepEqual(Object.keys(sent_data.context), ["id","root_id","url"]);
 			assert.ok(guid_re.test(sent_data.context.id), "Request ID is invalid. " + sent_data.context.id);
-			assert.equal(sent_data.context.root_id, "root_id");
+			assert.equal(sent_data.context.root_id, root_id);
 			assert.equal(sent_data.context.url, "http://www.ft.com/home/uk");
 
 			// User
-			assert.deepEqual(Object.keys(sent_data.user), ["user_id"]);
+			assert.deepEqual(Object.keys(sent_data.user), ["passport_id","ft_session","user_id"]);
```
rowanbeentje: I think this was a bug with the old core track implementation.  The documentation says "o-tracking will [...] Automatically pickup ftsession from cookies for you", and I don't think there's an expectation that setting a single user key will unset the defaults, unlike the other arrays.

rowanbeentje:
Although this is technically a major change because it removes the setRootID support from an "external" core singleton, I vote minor change.  Not fussed though.

src/javascript/core/user.js
```diff
@@ -8,6 +8,7 @@ let store;
 const defaultUserConfig = {
 	storage: 'cookie',
 	name: 'spoor-id',
+	nameOverride: 'spoor-id',
 	value: null,
 	domain: (document.URL.match(/^https?\:\/\/([^\/?#]+)(?:[\/?#]|$)/i)[1].indexOf('ft.com') > -1 ? 'ft.com' : null)
```
JakeChampion: Should this use `location.hostname` instead of `(document.URL.match(/^https?\:\/\/([^\/?#]+)(?:[\/?#]|$)/i)[1]`?

JakeChampion: This could also use `String.prototype.includes` instead of `.indexOf` as it is included in the default polyfill bundle -- https://polyfill.io/v2/docs/features/#String_prototype_includes

src/javascript/core.js
JakeChampion: Have you found any code which did consume the core object separately?

src/javascript/core/user.js
rowanbeentje: Shall I do that in a separate PR?

Agreed on using built-in instead of a regex.  It would also currently match on "microsoft.com"...

JakeChampion: Exactly 😂 . Separate PR is good with me 👌

src/javascript/core.js
rowanbeentje: Not based on going through https://github.com/search?q=org%3AFinancial-Times+o-tracking+core&type=Code .  Stash string matching with hyphens is terrible so I gave up going through the 1000+ search results there...

JakeChampion: I'm tempted to say _minor_ instead of _major_.

adambraimbridge: Perhaps 'root_id' was originally there to support testing? I'm completely clutching at straws with that guess.

rowanbeentje: Plausible, yes!  I like the way it returns it on generation though - better way of testing it...

rowanbeentje:
@chrisbrownuk be aware that as this gets rolled out, there may be a slight change in user data, but it should always be an improvement (previously if a custom user field was set, the default ft_session etc keys were overwritten rather than added to).

PR #158, opened 2017-03-15
Improve user domain matching regex, and remove migrate_across_domains now it's no longer needed

rowanbeentje:
I left the ft.com match as a regex as it's by far the simplest implementation, but fixed it!


rowanbeentje:
PS can anyone think of a better name than "user" for this class... it's closer to "device", but not quite...

PR #159, opened 2017-03-20
Bump the version number

rowanbeentje:
Oops.

(Can this be automated?  Even just adding it just to package.json would be an improvement - but then require'ing it on the client side may pull in the whole package, not idea.  Do you do this in other apps at the moment?)


JakeChampion:
Definitely can make it part of a build process.

PR #161, opened 2017-03-23
Stop config-less init

markstephens:
- If o-tracking initialises without config, it is useless
- The auto-init from the dom event keeps racing infront of my manual init with config
- This fix aims to avoid that race and discard the init coming from the dom-event
- Obvs, if the declaritive config is set, then it will still work with `domcontentloaded`


markstephens:
I've tried this on a website and it fixes the problem I'm seeing.

PR #164, opened 2017-09-27
Tp/removeid

tom-parker:


tavvy:
👍 🚢

notlee:
@tom-parker is this still something you would like merged?

tom-parker:
@notlee ah, yes I think so. Sorry, this one slipped my mind

notlee:
Cool. Out of interest why remove the id?

notlee:
@tom-parker Ah, second question (bare with me I'm new to o-tracking), does this not constitute a breaking change if someone is relying on the id for their tracking?

tom-parker:
I don't think this is a breaking change, from memory it was actually a bug. On click events I think `context.id` actually contains the id of the event. In circumstances where there was a id attribute on some clickable element, we'd end up overriding the id of the click event with the id attribute of the element if that makes sense.

notlee:
👍  :shipit:

markstephens:
`context.id` should be a unique ID for the event.
If you want to store another ID somewhere, it needs to go somewhere else.
Looking at Tom's changes, I think the code looks good to reflect this.

JakeChampion:
After merging in the latest changes from master, it looks like this work was done in a different pull-request.

PR #166, opened 2017-10-04
Add the tests which are not being run

JakeChampion:


JakeChampion:
@tom-parker Looks like these tests never ran, I've made them run and they look to fail.

wheresrhys:
fixed elsewhere

PR #172, opened 2017-11-14
info in urls now fixed

wheresrhys:
I was building the functionality aroudn the structure of the data fixture used in tests

It turns out the fixture was not an accurate picture of reality


tom-parker:
👍

PR #179, opened 2018-01-26
Pass secure token under a different attribute

tom-parker:
I'm hoping we can get this out ASAP. The previous change meant that the secure token was still being passed under the `ft_session` attribute. Unfortunately, membership have a different endpoints for validation of secure and non secure tokens `membership/sessions/s/` vs `membership/sessions/`

The previous change meant we were hitting the non secure endpoint with secure tokens and it was failing due to the invalid token type. This had the knock on affect of producing bad data at both Spoor and Ammit. 

I will make the relevant changes to use `ft_session_s` when available so hopefully we can ween ourselves off `ft_session`

@missamynicholson @rowanbeentje @Financial-Times/origami-core

### DIFF
--- a/docs/event-parameters/page-view.md
+++ b/docs/event-parameters/page-view.md
@@ -316,7 +316,7 @@ oTracking.init({
 
 - Key: `user.ft_session`
 - Required: no **yes, if available**
-- Default: *FTSession_s cookie*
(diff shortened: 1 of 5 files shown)


rowanbeentje:
Eeek, sorry.

PR #180, opened 2018-01-29
Allow cookie domain to be passed in via config

tom-parker:
This allows the cookie domain to be passed in via config.

Currently, if we're not on ft.com the cookie defaults to the current domain which is fine in most cases unless we're moving across subdomains. The spoor-id on foo.example.com will be different from bar.example.com which isn't desirable in some cases.

### DIFF
--- a/src/javascript/core/user.js
+++ b/src/javascript/core/user.js
@@ -24,10 +24,14 @@ const Store = require('./store');
  * @param {String} value The value of a userID to use if one is not stored
  * @return {String} - The user ID if present, or a generated UID if not
  */
-function init(value) {
-	store = new Store(defaultUserConfig.name, defaultUserConfig);
+function init(value, cookieDomain) {
+	if ( cookieDomain ) {
+		defaultUserConfig.domain = cookieDomain;
+	}
 
+	store = new Store(defaultUserConfig.name, defaultUserConfig);
 	let id = store.read();
+
 	if (!id) {
 		id = value;
 	}
--- a/src/javascript/core/user.js
+++ b/src/javascript/core/user.js
@@ -22,6 +22,7 @@ const Store = require('./store');
  * Init
  *
  * @param {String} value The value of a userID to use if one is not stored
+ * @param {String} cookieDomain The domain that should be used to cookie te user
  * @return {String} - The user ID if present, or a generated UID if not
  */
 function init(value, cookieDomain) {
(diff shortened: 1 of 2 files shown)


src/javascript/core/user.js
```diff
@@ -24,10 +24,14 @@ const Store = require('./store');
  * @param {String} value The value of a userID to use if one is not stored
  * @return {String} - The user ID if present, or a generated UID if not
```
rowanbeentje: Add as a `@param` here as well if that's the pattern so far?

rowanbeentje: (I also feel that perhaps it'd be more useful to pass the whole `config` object in here for future use...?)

rowanbeentje:
Makes sense and I hadn't realised this wasn't currently the case.  Will this have any adverse impact on the analytics side in terms of analysing journeys, or is this only a good thing?

(Seems like generally we should be using `.ft.com`!)

src/javascript/core/user.js
tom-parker: Yes this would probably benefit from being the whole config. I think throughout the rest of the code base, config is handled via the settings file. Unfortunately in the current iteration the user / store gets initialised before the config is saved to the settings file and so this change would require a bigger overhaul.

tom-parker:
This shouldn't make any difference from the ft.com side of things as everything is just defaulting to `.ft.com`. It's really needed for scout.asia where we go from `login.scout.asia` to `app.scout.asia` and we don't carry the cookies across those subdomains.

src/javascript/core/user.js
```diff
@@ -22,12 +22,17 @@ const Store = require('./store');
...
  *
  * @param {String} value The value of a userID to use if one is not stored
+ * @param {String} cookieDomain The domain that should be used to cookie te user
  * @return {String} - The user ID if present, or a generated UID if not
  */
-function init(value) {
-	store = new Store(defaultUserConfig.name, defaultUserConfig);
+function init(value, cookieDomain) {
+	if ( cookieDomain ) {
+		defaultUserConfig.domain = cookieDomain;
```
notlee: Is there a reason to manually pass the domain vs. calling `window.location.hostname`?
To hopefully answer my own question: Because `www.ft.com` would be unhelpful?

notlee: Sorry I missed the point, and this https://github.com/Financial-Times/o-tracking/blob/fcbb8b62bb80350b9c815f60bcc47bf8371548f2/src/javascript/core/user.js#L15

tom-parker: Yep, so this change is really for o-tracking usage outside of ft.com, in scout.asia for example where this default isn't desirable.

PR #183, opened 2018-03-26
Using `data-o-tracking-skip-queue` for skipping the queue and firing a link click straight away

adambraimbridge:
@dapenguin Check this shizzle

### DIFF
--- a/src/javascript/events/click.js
+++ b/src/javascript/events/click.js
@@ -27,10 +27,11 @@ const eventPropertiesToCollect = [
 
 // Trigger the event tracking
 const track = eventData => {
-	const href = eventData.context.domPathTokens[0].href || null;
+	const firstDomPathToken = eventData.context.domPathTokens[0];
+	const href = firstDomPathToken.href || null;
+	const skipQueue = firstDomPathToken['data-o-tracking-skip-queue'] === 'true' || false;
 	const isInternal = href && href.indexOf(window.document.location.hostname) > -1;
-	const skipQueue = eventData.context.skipQueue || false; //OMG REALLY??? ¯\_(ツ)_/¯ 
-	
+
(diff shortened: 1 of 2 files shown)


tom-parker:
What's the use case for this one?

src/javascript/events/click.js
```diff
@@ -27,10 +27,12 @@ const eventPropertiesToCollect = [
 
 // Trigger the event tracking
 const track = eventData => {
-	const href = eventData.context.domPathTokens[0].href || null;
+	const firstDomPathToken = eventData.context.domPathTokens[0];
+	const href = firstDomPathToken.href || null;
+	const skipQueue = firstDomPathToken['data-o-tracking-skip-queue'] === 'true' || false;
```
adambraimbridge: It might be worth adding a `.toLowerCase()` ... I'm not sure if that's already taken care of in the `domPathTokens` array.

``` javascript
const skipQueue = firstDomPathToken['data-o-tracking-skip-queue'].toLowerCase() === 'true' || false;
```

Side note: If you wanted to be a bit js-hipster about it, you could do this: (I think)
``` javascript
const skipQueue = !!(firstDomPathToken['data-o-tracking-skip-queue'] === 'true');
```

wheresrhys:
There's an ongoing debate about dropping the queue for all clicks. I started it a while ago but then dropped the ball. I'll try to wake it up again.

src/javascript/events/click.js
dapenguin: I'd have to check to make sure `firstDomPathToken['data-o-tracking-skip-queue']` exists before I can use `toLowerCase()`, but I'll add that in.

dapenguin:
@wheresrhys Is that something I should wait for? I've got a couple of bugs waiting on this.

PR #184, opened 2018-08-16
Do not read the location object on importing of the file

JakeChampion:
defaultUserConfig is only used within the init function. Moving the defaultUserConfig to be within the init function means that the user.js module will no longer try to read the location object upon being imported, this will help consumers of the module who are trying to use it in a server context where this is no location object present such as Node.JS


JakeChampion:
Yes @notlee

PR #188, opened 2018-09-05
IE11 fix

markstephens:
With this update the aim is to capture some events we've realised we're missing from IE11.

The changes mean o-tracking will fall back to image requests for IE11 when the AJAX version fails.
I have also added the transport method used to all requests so we can get stats on this.

This chart shows we are missing about 3% of events by not using Image tags (it's a subtraction of the IE11 test events from normal events, the pink area shows additional events we haven't been capturing)
screen shot 2018-09-05 at 10 31 38

But, this chart shows Image tags by themselves are not reliable enough, so we need to use the combination of AJAX and Image to try and get the most reliable result. (The image line is generally about 3%-5% below the AJAX line.)
screen shot 2018-09-05 at 10 31 45

### DIFF
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -7,7 +7,7 @@ const settings = require('./settings');
 const utils = require('../utils');
 const Queue = require('./queue');
 const transports = require('./transports');
-const ie11 = function () { return !!window.MSInputMethodContext && !!document.documentMode };
+const ie11 = function () { return !!window.MSInputMethodContext && !!document.documentMode; };
 /**
  * Default collection server.
  */
@@ -80,9 +80,9 @@ function sendRequest(request, callback) {
 			// If IE11 XHR error, try using image method
 			if (ie11() && transport.name === 'xhr') {
 				let image_method = transports.get('image')();
-				request.system.transport = [request.system.transport,image_method.name].join('-') // Append image label to transport value
+				request.system.transport = [request.system.transport,image_method.name].join('-'); // Append image label to transport value
 				image_method.send(url, JSON.stringify(request));
-				image_method.complete(function () {	
+				image_method.complete(function () {
 					if (callback) {
 						callback();
 					}
--- a/src/javascript/core/send.js
+++ b/src/javascript/core/send.js
@@ -7,7 +7,7 @@ const settings = require('./settings');
 const utils = require('../utils');
 const Queue = require('./queue');
 const transports = require('./transports');
-const ie11 = function () { return !!window.MSInputMethodContext && !!document.documentMode; };
+const isIe11 = function () { return !!window.MSInputMethodContext && !!document.documentMode; };
 /**
  * Default collection server.
  */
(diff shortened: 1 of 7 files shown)


test/core/send.test.js
```diff
@@ -223,6 +226,41 @@ describe('Core.Send', function () {
...
+			window.Image = sinon.stub().returns(dummyImage);
+
+			server.respondWith([500, { "Content-Type": "plain/text", "Content-Length": 5 }, "NOT OK"]);
+			Send.addAndRun(request);
+			server.respond();
+			// Respond OK for the image
+			server.respondWith([200, { "Content-Type": "plain/text", "Content-Length": 2 }, "OK"]);
+			server.respond();
+
+			console.log(dummyImage);
```
rowanbeentje: Was this meant to still be here?

Otherwise this is adding a chunk of complexity to the callback path but looks good and seems worth it for 3-5% of calls!

markstephens: good spot, i will remove the logging.

Yes, it's a bit busy on the callback area, but I also think worth it.

src/javascript/core/send.js
```diff
@@ -7,6 +7,7 @@ const settings = require('./settings');
 const utils = require('../utils');
 const Queue = require('./queue');
 const transports = require('./transports');
+const ie11 = function () { return !!window.MSInputMethodContext && !!document.documentMode; };
```
wheresrhys:
```
const isIe11 = !!window.MSInputMethodContext && !!document.documentMode;
```

 - execute once on startup and clearer variable name

src/javascript/core/send.js
```diff
@@ -75,15 +77,27 @@ function sendRequest(request, callback) {
...
-			queue.add(request).save();
-
-			utils.broadcast('oErrors', 'log', {
-				error: error.message,
-				info: { module: 'o-tracking' }
-			});
+			// If IE11 XHR error, try using image method
+			if (ie11() && transport.name === 'xhr') {
+				let image_method = transports.get('image')();
+				request.system.transport = [request.system.transport,image_method.name].join('-'); // Append image label to transport value
```
wheresrhys: I would put this comment above the line for legibility. Also... It doesn't explain why the image label is assigned

src/javascript/core/transports/send-beacon.js
```diff
@@ -6,6 +6,7 @@ module.exports = function () {
 		rejecter = reject;
 	});
 	return {
+		name: 'send-beacon',
```
wheresrhys: why not `sendBeacon`?

src/javascript/core/send.js
markstephens: I've updated the variable name. Having within a function makes the tests pass.

src/javascript/core/send.js
markstephens: Updated to be clearer

src/javascript/core/transports/send-beacon.js
markstephens: I don't know! Changed it!

PR #190, opened 2018-11-05
Image uses image

markstephens:
This is a test to see if IE is happier with an image response for Image transport rather than the normal html response from /ingest

See example traffic from a customer
screen shot 2018-11-05 at 12 33 51
It's recognising the AJAX method won't work and falling back to Images, which is great, but then also see the images don't make it through and are aborted.

### DIFF
--- a/src/javascript/core/transports/image.js
+++ b/src/javascript/core/transports/image.js
@@ -6,6 +6,7 @@ module.exports = function () {
 	return {
 		name: 'image',
 		send: function (url, data) {
+			url = url.replace('https://spoor-api.ft.com/ingest', 'https://spoor-api.ft.com/px.gif');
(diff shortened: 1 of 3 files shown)


src/javascript/core/transports/image.js
```diff
@@ -6,6 +6,7 @@ module.exports = function () {
 	return {
 		name: 'image',
 		send: function (url, data) {
+			url = url.replace('https://spoor-api.ft.com/ingest', 'https://spoor-api.ft.com/px.gif');
```
markstephens: I limited it to rewriting the ft.com ingest point only (https://spoor-api.ft.com/ingest) as I didn't want to disrupt NAR in any way (https://github.com/Financial-Times/o-tracking-builder-service/blob/master/views/products/nar.pug#L4)

PR #196, opened 2019-02-08
use ESM insteadof CJS

JakeChampion:


src/javascript/core/transports/image.js
```diff
@@ -20,3 +20,6 @@ module.exports = function () {
 		}
 	};
 };
+
+export default image;
+export { image };
```
chee: this is tricky, these are different between commonjs, commonjs w/ interop and es6

not sure what the best option is here

chee: here, the codeblocks contain the value of the import.

i might be wrong about this, but i think this results in the following:

### commonjs without noInterop `let image = require("./image")`

```js
{image: image, default: image}
```

### commonjs with noInterop `let image = require("./image")`

```js
{default: {image: image, default: image}}
```

### es6 `import image from "./image"`
```
image
```

### es6 `import {image} from "./image"`
```
image
```

chee: i may well be wrong and someone sould probably try it out

notlee:
Closing for now, until a future breaking release: https://github.com/Financial-Times/o-tracking/issues/207

PR #199, opened 2019-02-25
Add impression(component:view) tracking

fenglish:
Component view events will be fired with elements set `data-o-tracking-view` attribute or elements collected by a custom selector in `opts`. They will populate a `domPathTokens` as well as Click events. You could pass a custom category, and also a callback[`getContextData`] to collect data for events.  

:warning: This feature requires `window.IntersectionObserver` in order to track the events  
:warning: `getContextData` should be a function which returns `{Object}`

```js
const opts = {
	category: 'audio', // default: 'component'
	selector: '.o-teaser__audio', // default: '[data-o-tracking-view]'
	getContextData: () => {  // default: null
		return {
			contentId: '1234',
			type: 'audio',
			subtype: 'podcast',
		 	component: 'teaser'
		};
	}
}

oTracking.view.init(opts);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ function oTrackinginit() {
     oTracking.page({...config...});
     // Clicks.
     oTracking.click.init(category);
+    // Components/Elements views.
+    oTracking.view.init({...opts...});
 }
 ```
 
@@ -79,6 +81,8 @@ if (cutsTheMustard) {
     oTracking.page({...config...});
     // Clicks
     oTracking.click.init(category);
+    // Components/Elements views.
+    oTracking.view.init({...opts...});
 }
 ```
 
@@ -188,6 +192,25 @@ o-tracking will
 * Automatically pickup ftsession from cookies for you.
 * Page events automatically pick up the url and the referrer.
 * Click events [initalised as above](#usage), will populate a `domPathTokens` property. If the clicked element has the `data-trackable` attribute, sibling elements will also be included within `domPathTokens`.
+* View events are fired for elements with the `data-o-tracking-view` attribute by default, unless `o-tracking`'s `selector` option is configured. Like click events, view events populate a `domPathTokens` property. To collect data for events, set the `category` option, or provide a callback[`getContextData`]  
+_Note:_ This feature requires `window.IntersectionObserver` in order to track the events  
+_Note:_ `getContextData` should be a function which returns `{Object}`. It accepts the viewed element as an argument  
+	```js
+	const opts = {
+		category: 'audio', // default: 'component'
+		selector: '.o-teaser__audio', // default: '[data-o-tracking-view]'
+		getContextData: (el) => {  // default: null
+			return {
+				contentId: el.getAttribute('data-id'),
+				type: 'audio',
+				subtype: 'podcast',
+			 	component: 'teaser'
+			};
+		}
+	}
+
+	oTracking.view.init(opts);
+	```
 
 Events are decorated with config values you pass in via `init()` or `updateConfig()` if they are part of `context`, `user`, or `device` objects. Values passed in as `context` to individual events will override context values from init.
 
--- a/src/javascript/events/component-view.js
+++ b/src/javascript/events/component-view.js
@@ -1,0 +1,87 @@
+/*global module, require */
+'use strict'; // eslint-disable-line strict
+
+const Core = require('../core');
+const getTrace = require('../../libs/get-trace');
+const utils = require('../utils');
+
+const TRACKING_ATTRIBUTES = [
+	'contentId',
+	'type',
+	'subtype',
+	'component',
+];
+
+const decorateEventData = (eventData, viewedEl, opts) => {
+
+	const { trace, customContext } = getTrace(viewedEl);
+	let context;
+
+	if (opts.getContextData) {
+		if (typeof opts.getContextData !== 'function') {
+			throw new Error('opts.getContextData is not a function');
+		}
+
+		const dataBeforeWhitelist = opts.getContextData(viewedEl);
+
+		if (typeof dataBeforeWhitelist !== 'object') {
+			throw new Error('opts.getContextData function should return {Object}');
+		}
+
+		context = utils.whitelistProps(dataBeforeWhitelist, TRACKING_ATTRIBUTES);
+	} else {
+		context = {};
+	}
+
+	context.domPathTokens = trace;
+	context.url = window.document.location.href || null;
+
+	utils.assignIfUndefined(customContext, context);
+
+	eventData.context = context;
+};
+
+/**
+ * Listen for view events.
+ *
+ * @alias view#init
+ * @param {Object} opts - To set custom category[String], selector[String], getContextData[Function]
+ * @return {undefined}
+ */
+const init = (opts = {}) => {
+	if(!window.IntersectionObserver) {
+		return;
+	}
+
+	const selector = opts.selector || '[data-o-tracking-view]';
+	const elementsToTrack = [...document.querySelectorAll(selector)];
+
+	if (!elementsToTrack.length) {
+		console.warn('o-tracking: Unable to track element view events as "window.IntersectionObserver" is not supported.');
+		return;
+	}
+
+	function onChange (changes) {
+		changes.forEach(change => {
+			if(change.isIntersecting || change.intersectionRatio > 0) {
+				const eventData = {
+					action: 'view',
+					category: opts.category || 'component'
+				};
+				const viewedEl = change.target;
+
+				decorateEventData(eventData, viewedEl, opts);
+				Core.track(eventData);
+				this.unobserve(viewedEl);
+			}
+		});
+	}
+
+	const observer = new IntersectionObserver(onChange, { threshold: [ 1.0 ] });
+
+	elementsToTrack.forEach(el => observer.observe(el));
+};
+
+module.exports = {
+	init: init
+};
--- a/src/javascript/utils.js
+++ b/src/javascript/utils.js
@@ -214,6 +214,47 @@ function getValueFromJsVariable(str) {
 	return test !== '' ? test : null;
 }
 
+/**
+ * Trim strings
+ * @param {String} str - The string to trim.
+ * @return {String} The trimmed string.
+ */
+function sanitise (str) {
+	return typeof str === 'string' ? str.trim() : str;
+}
+
+/**
+ * Assign the subject value if the target properties are undefined
+ * @param {Object} subject - assign the value
+ * @param {Object} target - be assigned the value
+ * @return
+ */
+function assignIfUndefined (subject, target) {
+	for (const prop in subject) {
+		if (!target[prop]) {
+			target[prop] = subject[prop];
+		} else {
+			console.warn(`You can't set a custom property called ${prop}`);
+		}
+	}
+}
+
+/**
+ * Whitelist props
+ * @param {Object} props - An object whose props need to be whitelisted
+ * @param {Array} list - A list for whitelisting
+ * @return
+ */
+function whitelistProps (props, list) {
+	return list.reduce(
+		(acc, propName) => Object.assign(
+			{},
+			acc,
+			props[propName] ? { [propName]: props[propName] } : undefined
+		),
+		{}
+	);
+}
 
 /**
  * Utilities.
@@ -233,5 +274,8 @@ module.exports = {
 	triggerPage: triggerPage,
 	getValueFromCookie: getValueFromCookie,
 	getValueFromUrl: getValueFromUrl,
-	getValueFromJsVariable: getValueFromJsVariable
+	getValueFromJsVariable: getValueFromJsVariable,
+	sanitise: sanitise,
+	assignIfUndefined: assignIfUndefined,
+	whitelistProps: whitelistProps,
 };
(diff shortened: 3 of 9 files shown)


src/javascript/events/component-view.js
```diff
@@ -0,0 +1,58 @@
...
+ * @alias view#init
+ * @param {Object} opts - To set custom category, elementsToTrack, callback to collect data for the event
+ * @return {undefined}
+ */
+const init = (opts = {}) => {
+	if(!window.IntersectionObserver) {
+		return false;
+	}
+
+	const elementsToTrack = opts.elementsToTrack || [...document.querySelectorAll('[data-o-tracking-view]')];
```
robsquires: I'm wondering if passing a selector here might be simpler for the applications using this + possibly less prone to misconfiguration (e.g. what if you don't pass an array of HTMLElements) ? 🤔

README.md
```diff
@@ -188,6 +192,16 @@ o-tracking will
 * Automatically pickup ftsession from cookies for you.
 * Page events automatically pick up the url and the referrer.
 * Click events [initalised as above](#usage), will populate a `domPathTokens` property. If the clicked element has the `data-trackable` attribute, sibling elements will also be included within `domPathTokens`.
+* Component view events will be fired with elements set `data-o-tracking-view` attribute or passed as elementsToTrack[array] in `opts`. They will populate a `domPathTokens` as well as Click events. You could pass custom category and also callback to collect data for events.  
+	```js
+	const opts = {
+		category: 'custom', // default: 'component'
+		elementsToTrack: [ element, element, ... ], // default: elements set data-o-tracking-view
+		callback: () => {} // default: null
```
robsquires: Wondering if we could make the name of this more explicit, something like ... `getContextForComponent`  🤔 ?

test/events/component-view.test.js
```diff
@@ -0,0 +1,140 @@
...
+			targetComponent = document.createElement('div');
+			targetComponent.setAttribute('data-o-tracking-view', true);
+			targetComponent.innerHTML = 'Default component view elements';
+			document.body.appendChild(targetComponent);
+
+			componentView.init();
+			viewed(targetComponent);
+		});
+
+		it('should track an event for a component view ', () => {
```
robsquires: Would it be worth adding a case for `should only track an event for a component view once` ?

test/events/component-view.test.js
```diff
@@ -0,0 +1,140 @@
...
+			targetComponent.innerHTML = 'Custom component view elements';
+			targetComponent.setAttribute(targetAttribute, true);
+			targetComponent.setAttribute(idAttribute, id);
+			document.body.appendChild(targetComponent);
+
+			const opts = {
+				category,
+				elementsToTrack: [...document.querySelectorAll(`[${targetAttribute}]`)],
+				callback: (el) => {
+					optsCallbackSpy(el);
```
robsquires: Minor thing, but given you're asserting on the data returned by this function, you could probably eschew spying on this function as well.

src/javascript/events/component-view.js
```diff
@@ -0,0 +1,58 @@
...
+
+	function onChange (changes) {
+		changes.forEach(change => {
+			if(change.isIntersecting || change.intersectionRatio > 0) {
+				const eventData = {
+					action: 'view',
```
benbarnett: I know we sort of settled on this now but I'm wondering if `component:seen` might be clearer........ what do you think? No worries if we're too far down the line!

fenglish: @benbarnett The business analyst side wants to keep consistent for the action.
`view` is already used for impressions. e.g. comment:view.

test/events/component-view.test.js
fenglish: @robsquires I thought That is a IntersectionObserver side test. It might be a bit tricky to test that?
It means `mockIntersectionObserver` has to have the code for `unobserve`.

There is a line to check `unobserve` is called. Is it not enough?
`assert.equal(unobserveSpy.calledOnce, true, "IntersectionObserver unobserved target");`

robsquires: Could you fire `viewed(targetComponent);` twice?

src/javascript/events/component-view.js
```diff
@@ -0,0 +1,67 @@
...
+const Core = require('../core');
+const getTrace = require('../../libs/get-trace');
+const utils = require('../utils');
+
+const decorateEventData = (eventData, viewedEl, opts) => {
+
+	const { trace, customContext } = getTrace(viewedEl);
+	let context = typeof opts.getContextData === 'function' ? opts.getContextData(viewedEl) : {};
+
+	if (typeof context !== 'object') {
```
benbarnett: Is this block needed?

I guess this is guarding against consumers providing an invalid callback function, i.e. one that doesn't return an object. If that is the case then I think this could probably make people think they are using this method correctly, when in fact it should be erroring...

src/javascript/utils.js
```diff
@@ -214,6 +214,30 @@ function getValueFromJsVariable(str) {
 	return test !== '' ? test : null;
 }
 
+/**
+ * Trim strings
+ * @param {String} str - The string to trim.
+ * @return {String} The trimed string.
```
benbarnett: typo: `trimmed`

src/javascript/utils.js
```diff
@@ -233,5 +257,7 @@ module.exports = {
 	triggerPage: triggerPage,
 	getValueFromCookie: getValueFromCookie,
 	getValueFromUrl: getValueFromUrl,
-	getValueFromJsVariable: getValueFromJsVariable
+	getValueFromJsVariable: getValueFromJsVariable,
+	sanitise: sanitise,
+	assignIfUndefined: assignIfUndefined
```
benbarnett: Looks like we have a mix of trailing commas and no trailing commas (i.e. on `src/libs/get-trace.js`)

If it this is true for the rest of the repo perhaps it could be cleaned up in a separate PR.

test/events/component-view.test.js
```diff
@@ -0,0 +1,208 @@
+/*global require, describe, it, before, after, sinon, document */
+require('../setup');
+const assert = require("assert");
+const settings = require("../../src/javascript/core/settings");
```
benbarnett: These imports typically use single quotes

src/javascript/events/component-view.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Listen for view events.
+ *
+ * @alias view#init
+ * @param {Object} opts - To set custom category[String], selector[String], getContextData[Function]
+ * @return {undefined}
+ */
+const init = (opts = {}) => {
+	if(!window.IntersectionObserver) {
```
benbarnett: Is it worth adding somewhere to the readme that this feature requires `window.IntersectionObserver` in order to track the events? https://caniuse.com/#feat=intersectionobserver

In FT.com we are loading the polyfill in n-ui, so not a problem 😀

src/javascript/events/component-view.js
```diff
@@ -0,0 +1,67 @@
...
+'use strict'; // eslint-disable-line strict
+
+const Core = require('../core');
+const getTrace = require('../../libs/get-trace');
+const utils = require('../utils');
+
+const decorateEventData = (eventData, viewedEl, opts) => {
+
+	const { trace, customContext } = getTrace(viewedEl);
+	let context = typeof opts.getContextData === 'function' ? opts.getContextData(viewedEl) : {};
```
robsquires: I'm wondering if it's worth running the object returned from `getContextData` through a whitelist like [o-audio](https://github.com/Financial-Times/o-audio/blob/master/src/js/tracking.js#L46). In practice these attributes need to be agreed with the data team + would provide some free documentation as to what can be returned. What ya think?

src/javascript/utils.js
fenglish: 🙈

test/events/component-view.test.js
fenglish: Oh thank you!

src/javascript/events/component-view.js
fenglish: Oh,That is a good point. I'm going to add the explanation.

src/javascript/events/component-view.js
fenglish: I see. I will keep the block because if the function returns String, it won't throw an error, just will mess up the data detail. So I will use the block to throw an error.

src/javascript/events/component-view.js
fenglish: I see! I will do that. Thank you!

src/javascript/events/component-view.js
benbarnett: Sounds sensible!

test/events/component-view.test.js
```diff
@@ -0,0 +1,226 @@
...
+	context: {
+		product: 'desktop'
+	},
+	user: {
+		user_id: '123456'
+	},
+};
+
+let targetComponent;
+let error;
```
benbarnett: I think this is good as it is, but do you think this would be marginally clearer if it was called `errorThrown`?

test/utils.test.js
```diff
@@ -70,4 +70,42 @@ describe('Utils', function () {
...
+			'contentId',
+			'subtype',
+			'type',
+		];
+		const resultProps = { contentId: 1234, subtype: 'subtype' };
+		const testProps = Object.assign(
+			{},
+			resultProps,
+			{ type: undefined, name: 'name' }
+		);
```
benbarnett: Tiny thing again but you could refactor this to match the array style in the rest of this spec?

fenglish: 👍

test/events/component-view.test.js
fenglish: I agree!

src/javascript/events/component-view.js
JakeChampion: It should be added to the `origami.json.browserFeatures.required` array. --> https://origami.ft.com/docs/syntax/origamijson/

fenglish: @JakeChampion Thank you for telling me!

README.md
```diff
@@ -188,6 +192,25 @@ o-tracking will
 * Automatically pickup ftsession from cookies for you.
 * Page events automatically pick up the url and the referrer.
 * Click events [initalised as above](#usage), will populate a `domPathTokens` property. If the clicked element has the `data-trackable` attribute, sibling elements will also be included within `domPathTokens`.
+* Component view events will be fired with elements set `data-o-tracking-view` attribute or elements collected by a custom selector in `opts`. They will populate a `domPathTokens` as well as Click events. You could pass a custom category, and also a callback[`getContextData`] to collect data for events.  
+:warning: This feature requires `window.IntersectionObserver` in order to track the events
```
notlee:
```suggestion
_Note:_ This feature requires `window.IntersectionObserver` in order to track the events.
```

This is super trivial, but maybe we could say "note" rather than "warning" here, and below for `getContextData`? Warnings are scary 🙈

README.md
```diff
@@ -188,6 +192,25 @@ o-tracking will
...
 * Page events automatically pick up the url and the referrer.
 * Click events [initalised as above](#usage), will populate a `domPathTokens` property. If the clicked element has the `data-trackable` attribute, sibling elements will also be included within `domPathTokens`.
+* Component view events will be fired with elements set `data-o-tracking-view` attribute or elements collected by a custom selector in `opts`. They will populate a `domPathTokens` as well as Click events. You could pass a custom category, and also a callback[`getContextData`] to collect data for events.  
+:warning: This feature requires `window.IntersectionObserver` in order to track the events  
+:warning: `getContextData` should be a function which returns `{Object}`
+	```js
+	const opts = {
+		category: 'audio', // default: 'component'
+		selector: '.o-teaser__audio', // default: '[data-o-tracking-view]'
+		getContextData: () => {  // default: null
```
notlee: If I understand correctly, the callback function accepts the viewed element as an argument? It would be great to document that here.

src/javascript/events/component-view.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Listen for view events.
+ *
+ * @alias view#init
+ * @param {Object} opts - To set custom category[String], selector[String], getContextData[Function]
+ * @return {undefined}
+ */
+const init = (opts = {}) => {
+	if(!window.IntersectionObserver) {
```
notlee: It's good that this is optional to avoid a breaking change :) But perhaps a warning here would be good to help debug, if users expect to track views across older browsers?
```
console.warn('o-tracking: Unable to track element view events as "window.IntersectionObserver" is not supported.');
```

fenglish: 👍

README.md
fenglish: 👍

README.md
fenglish: 🙈

README.md
```diff
@@ -188,6 +192,25 @@ o-tracking will
 * Automatically pickup ftsession from cookies for you.
 * Page events automatically pick up the url and the referrer.
 * Click events [initalised as above](#usage), will populate a `domPathTokens` property. If the clicked element has the `data-trackable` attribute, sibling elements will also be included within `domPathTokens`.
+* Component view events will be fired with elements set `data-o-tracking-view` attribute or elements collected by a custom selector in `opts`. They will populate a `domPathTokens` as well as Click events. You could pass a custom category, and also a callback[`getContextData`] to collect data for events.
```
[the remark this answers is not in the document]
fenglish: I see. Thank you for making them nicer!

PR #200, opened 2019-03-15
component:view tracking: Modify contentId => componentContentId

fenglish:
Impressions on article pages having both **contentID**(the component's content id) and **rootContentID**(the article's content id) is potentially confusing. We rename **contentID** to **componentContentID** so it is very clear what that applies to.

### DIFF
--- a/src/javascript/events/component-view.js
+++ b/src/javascript/events/component-view.js
@@ -6,7 +6,7 @@ const getTrace = require('../../libs/get-trace');
 const utils = require('../utils');
 
 const TRACKING_ATTRIBUTES = [
-	'contentId',
+	'componentContentId',
 	'type',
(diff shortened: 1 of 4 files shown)


src/javascript/events/component-view.js
```diff
@@ -6,7 +6,7 @@ const getTrace = require('../../libs/get-trace');
 const utils = require('../utils');
 
 const TRACKING_ATTRIBUTES = [
-	'contentId',
+	'componentContentId',
```
JakeChampion: Is this a breaking change?

fenglish: @JakeChampion It is, but this component-view hasn't been used anywhere yet...

PR #201, opened 2019-04-08
Bump ftdomdelegate

liamkeaton:


i-like-robots:
I did not anticipate this issue when re-naming the module but I suppose it is something we need to push through. Part of the work done to publish Origami components to npm was to add mappings for packages with different names, one of those is `ftdomdeletegate` => `dom-delegate`.

https://github.com/Financial-Times/origami-component-converter/blob/4380b360f871728dc8ce04f6eee587ff8e61be76/config/mappings.js

notlee:
I think we can make this non-breaking by using a range, so going to close this one for now: 
https://github.com/Financial-Times/o-tracking/compare/ftdomdelegate-bump

PR #203, opened 2019-04-09
Expose getRootID() on the oTracking object

alexflorisca:
We require Markets Data to add a rootID to the page and pass it into oAds. They don't use n-ui but they do use oTracking, from the build service. This PR enables them to get a rootId for the markets data page.

### DIFF
--- a/test/main.test.js
+++ b/test/main.test.js
@@ -261,4 +261,8 @@ describe('main', function () {
 		assert.equal(data.context.marketing.banner_closed, true);
 		assert.equal(data.context.marketing.segid, '4321');
 	});
+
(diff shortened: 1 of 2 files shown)


test/main.test.js
```diff
@@ -261,4 +261,8 @@ describe('main', function () {
 		assert.equal(data.context.marketing.banner_closed, true);
 		assert.equal(data.context.marketing.segid, '4321');
 	});
+
+	it('should have getRoodID() as an accessible method', () => {
```
rowanmanning:
```suggestion
	it('should have getRootId() as an accessible method', () => {
```

PR #204, opened 2019-05-07
Fix wrong warning for component view

fenglish:
I have added `console.warn` about `window.IntersectionObserve` in wrong place 😱


fenglish:
Thank you! @JakeChampion

Issue #195, opened 2019-01-17
Remove the version number from code


notlee:
The components version is specified in [`main.js` and `package.json`](https://github.com/Financial-Times/o-tracking/pull/194) and should be removed. 

Although it specifies `bower.json`, I also think it's fair to say it contradicts the [Origami specification](https://origami.ft.com/docs/component-spec/modules/).
>Must not include a version property. The version property is not needed and risks being out of sync with the repo tag

If there's a usecase for getting the version number from component JS, perhaps there's a way we can automate it across all our components.

JakeChampion:
o-ads uses genversion and release-it scripts to automate adding a version property -- https://github.com/Financial-Times/o-ads/pull/192/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R20

PR #221, opened 2019-11-13
Remove version from `package.json`. Update version in `main.js`

notlee:
Thie commit removes the version number from `package.json`
to align with the Origami spec (it's currently outdated anyway).

The version number is also set in `main.js` is incorrect. This commit
updates it for the new major. It's likely to fall out of date again.
We should consider removing it in a future version of o-tracking but
first need to understand if/how it's used.

https://github.com/Financial-Times/o-tracking/issues/195

### DIFF
--- a/main.js
+++ b/main.js
@@ -11,7 +11,7 @@ const core = require('./src/javascript/core');
  * The version of the tracking module.
  * @type {string}
  */
-const version = '1.4.2';
+const version = '2.0.0';
 /**
  * The source of this event.
  * @type {string}
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
 	"name": "o-tracking",
 	"description": "This module should be included on your product to make sending tracking requests to the Spoor API easy.",
-	"version": "1.4.2",
 	"main": "main.js",
 	"private": true,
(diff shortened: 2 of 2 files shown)


origamiserviceuser:
o-tracking bundle size difference from 1.7.3 to 3cfea4ef6f675e7909040f4d9fb884da1415a433

JakeChampion:
The version is sent with each request sent to spoor -- https://github.com/Financial-Times/o-tracking/blob/21e43d1fcb8d003c7e7e91ae9f396dc2e79764c6/src/javascript/core/send.js#L51

notlee:
> The version is sent with each request sent to spoor --
> 
> https://github.com/Financial-Times/o-tracking/blob/21e43d1fcb8d003c7e7e91ae9f396dc2e79764c6/src/javascript/core/send.js#L51

But I wonder if it's used 🤔 It can't hold up the major cascade, so let's leave for a future o-tracking focused work

origamiserviceuser:
o-tracking bundle size difference from 1.7.3 to 4dfc5bbef4c918881d8b3069709c4f230ccc7f9d
js: 3.9kb increase (0.73kb/gzip)

PR #222, opened 2019-11-13
Remove the `main` field from package.json.

notlee:
This is not allowed by the spec.
https://origami.ft.com/spec/v1/components/#npm


origamiserviceuser:
o-tracking bundle size difference from 1.7.3 to a465bd393a3cfd4809700648e81985edd6ab5b12

PR #225, opened 2019-11-22
Export the o-tracking init function.

notlee:
Many users are including o-tracking without a semver range:
`https://build.origami.ft.com/bundles/js?modules=o-tracking`

This makes o-tracking v2 compatible with v1 for their usecase via
the build service.


origamiserviceuser:
o-tracking bundle size difference from 2.0.0 to 8094cd44afa2932f8b80d9944a3a7d005d1815bb

PR #228, opened 2019-12-04
Export an object containing init from events/component-view

chee:
because that's what used to be exported:
https://github.com/Financial-Times/o-tracking/blob/v1.7.3/src/javascript/events/component-view.js#L85-L87

(the init function was never the main export)

### DIFF
--- a/src/javascript/events/component-view.js
+++ b/src/javascript/events/component-view.js
@@ -1,4 +1,3 @@
-
 import Core from '../core';
 import getTrace from '../../libs/get-trace';
 import utils from '../utils';
@@ -80,6 +79,6 @@ const init = (opts = {}) => {
 	elementsToTrack.forEach(el => observer.observe(el));
 };
 
-export default init;
+export default { init };
 export { init };
 
--- a/test/events/component-view.test.js
+++ b/test/events/component-view.test.js
@@ -84,7 +84,7 @@ describe('component:view', () => {
 			const text = 'component:view target for default props';
 
 			createTargetComponent(attributes, text);
-			componentView();
+			componentView.init();
 			viewed(targetComponent);
 		});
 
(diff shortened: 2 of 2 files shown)


src/javascript/events/component-view.js
```diff
@@ -80,6 +79,6 @@ const init = (opts = {}) => {
 	elementsToTrack.forEach(el => observer.observe(el));
 };
 
-export default init;
+export default { init };
```
JakeChampion: I guess this would constitute a breaking change? If so, maybe it would be better to update the documentation/migration-guide to mention that it no exports an object?

chee: the change to this *was* an undocumented breaking change, it has broken `next-search-page`. i think this constitutes a patch

chee: the migration guide says that what has changed between versions 1 and 2 is limited to the main interface https://github.com/Financial-Times/o-tracking/blob/master/MIGRATION.md

but that isn't true, because this change also took place and breaks next-search-page.

`modules.exports` was an object containing `init` before: https://github.com/Financial-Times/o-tracking/blob/v1.7.3/src/javascript/events/component-view.js#L85-L87

so when you accessed `oTracking.view` you would have an object containing a function called `init`, but now `oTracking.view` is a function

chee: the README also contains several mentions to `oTracking.view.init()` which is no longer possible:
https://github.com/Financial-Times/o-tracking/tree/v1.7.3#parameters

chee: you can see that we changed the tests to fit with the new api without documenting the api had been changed: https://github.com/Financial-Times/o-tracking/commit/a2de177e2dcc82edbcf94409c8357e9ba3a2287c#diff-3678da481657f896f0654a97774b6694L87-L88

origamiserviceuser:
o-tracking bundle size difference from 2.0.3 to 79acc2c85ac2920fc342d85eafd620bb8e0f64ad

PR #236, opened 2020-04-01
Stop using magic webpack imports

JakeChampion:
Stop using the magic imports for proclaim and sinon and pull them from the global because that's where Karma has added them. This is required for obt v10 to be released.


origamiserviceuser:
o-tracking bundle size difference from 2.0.7 to a5c8f42a8f9e4b7b3e4eb0f4a7f4e73cf5f82297

PR #260, opened 2020-07-14
Remove the makefile as every task/target in it does not work anymore

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to f10381666038a2c27812b083cac54d1bb1f70c52

PR #262, opened 2020-07-14
make version property contain the correct version

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 106324e3ca86d11f41d8bbecef53cd11af8dcfb1

Issue #115, opened 2016-02-15
Handle unserialisable JSON


wheresrhys:
`Circular reference in value argument not supported` gets thrown in IE. Should we
- catch and swallow the error
- catch the error and send something more descriptive using o-errors
- include a library which handles unserialisable JSON better, and send its output as a normal tracking event
- the above, but also add some properties to indicate there's an error
- let it be
- none of the above

JakeChampion:
I've contacted the data-platform team to see if Spoor supports circular references, which will help decide what o-tracking should do.

I've also listed my own thoughts to the potential solutions that were suggested 👍 

> `Circular reference in value argument not supported` gets thrown in IE

It is also thrown in the latest version of Chrome.

>Should we
> * catch and swallow the error

I don't think we should catch and swallow the error because that could give the team who is using o-tracking the impression that this specific piece of data is being tracked when in fact it is not.

> * catch the error and send something more descriptive using o-errors

We could do this but then the application which uses o-tracking may not be using o-errors so they would not get the error.

> * include a library which handles unserialisable JSON better, and send its output as a normal tracking event

I don't think we should handle circular references as JSON does not support them at all and each library solves circular references in their own way. If we wanted to support them, we would need to ensure that the Spoor pipeline uses the same circular reference parsing/stringifying system that o-tracking would use.

> * the above, but also add some properties to indicate there's an error

Same concern as the above question but would also potentially confuse the team using o-tracking because they would see the data would have reached Spoor correctly but see an error in their developer tools.

> * let it be

I'm leaning towards this but perhaps improving the error message to mention why we do not support circular references.

> * none of the above

JakeChampion:
@wheresrhys Do you have a real-world example of when an application would want circular references for tracking purposes? This issue is very old (2016) so I understand if you can't remember of an example.

JakeChampion:
The Spoor API uses [JSON.parse](https://github.com/Financial-Times/spoor-api/blob/1b784495d755a13710b2bb53270aadcf8ca9c450/server/api.js#L61), which means it does not understand circular references either.

I think because spoor-api does not understand circular references and no other requests have been made to support them, o-tracking should not support circular references either.

wheresrhys:
Can't remember sorry - was a long time ago. The fact I came across an error in IE hints that this was an actual real world problem though.

JakeChampion:
[...]

PR #263, opened 2020-07-15
When given a circular object to send to spoor, throw a custom error message that states circular objects are not supported 

JakeChampion:
Resolves https://github.com/Financial-Times/o-tracking/issues/115

After speaking with the Data team and Tom Parker, we have come to the conclusion that circular objects are not going to be supported in Spoor and that o-tracking should throw an error if given a circular object

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,6 @@
   ],
   "description": "Origami module for FT tracking.",
   "dependencies": {
-    "ftdomdelegate": "^4.0.0",
-    "is-circular": "tjmehta/is-circular#^1.0.2"
+    "ftdomdelegate": "^4.0.0"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -11,6 +11,7 @@
   ],
   "description": "Origami module for FT tracking.",
   "dependencies": {
-    "ftdomdelegate": "^4.0.0"
+    "ftdomdelegate": "^4.0.0",
+    "is-circular": "tjmehta/is-circular#^1.0.2"
   }
 }
--- a/src/javascript/utils.js
+++ b/src/javascript/utils.js
@@ -299,6 +299,54 @@ function findCircularPathsIn(rootObject) {
 }
 
 /**
+ * Used to find out whether an object contains a circular reference.
+ * @param {*} rootObject The object we want to search within for circular references
+ * @returns {bool} Returns true if a circular reference was found, otherwise returns false
+ */
+function containsCircularPaths(rootObject) {
+	// Used to keep track of all the values the rootObject contains
+	const traversedValues = new WeakSet();
+
+	/**
+	 *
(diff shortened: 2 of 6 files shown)


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 7a426ec3d00649e35bc1ce5d53c362e551ae0ac0
js: 5.62kb increase (1.91kb/gzip)

bower.json
```diff
@@ -11,6 +11,7 @@
   ],
   "description": "Origami module for FT tracking.",
   "dependencies": {
-    "ftdomdelegate": "^4.0.0"
+    "ftdomdelegate": "^4.0.0",
+    "is-circular": "tjmehta/is-circular#^1.0.2"
```
notlee, line 15: `tjmehta/is-circular` is quite a small package without many contributors. For security maybe we should implement directly in `o-tracking` or lock down to a specific version?

src/javascript/utils.js
```diff
@@ -253,6 +253,51 @@ function assignIfUndefined (subject, target) {
 	}
 }
 
+/**
+ * Used to find out all the paths which contain a circular reference.
+ * @param {*} rootObject The object we want to search within for circular references
+ * @returns {string[]} Returns an array of strings, the strings are the full paths to the circular references within the rootObject
+ */
+function findCircularPathsIn(rootObject) {
+	const traversedValues = new WeakSet();
```
notlee, line 262: Users would need a new polyfill for `WeakSet`so this is a breaking change

JakeChampion: I changed the base branch to merge into v3 instead

bower.json
JakeChampion: Sure thing, I coded one up myself but decided to use their version instead for some reason --
https://gist.github.com/JakeChampion/e1c593b67c59d0f75c7ca3c1cb86a21f

src/javascript/utils.js
notlee: cool :) don't forget to add it to `origami.json`!

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to cb743ab1919be546c1b6bc8529ffd46f36b643c4
js: 5.76kb increase (1.88kb/gzip)

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 080d5713e35ef0c4e29d462fff2a3092094f8a0e
js: 5.76kb increase (1.88kb/gzip)

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 93fba8280f151ba29548d35115de75a1050561c7
js: 5.48kb increase (1.78kb/gzip)

JakeChampion:
@notlee please take a look again, I've now updated the migration guide and origami.json.browserFeatures.required array and swapped out is-circular for our own implentation

PR #265, opened 2020-07-21
Fix some JSDoc return types

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to da1ccfdb97d0195f55f5c8d8ff1e239fc81b36a2
js: 5.48kb increase (1.78kb/gzip)

PR #267, opened 2020-07-21
 Remove setDomain and getDomain as only one domain is supported by o-tracking and spoor

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 6215633f315820070d824e3cb211d534abafb44f
js: 5.1kb increase (1.64kb/gzip)

PR #268, opened 2020-07-21
Move JSDoc for version and source to the Tracking.prototype

JakeChampion:
With this change, when users type in `tracking.version` or `tracking.source` it will let them know what the property is for


src/javascript/tracking.js
```diff
@@ -0,0 +1,250 @@
...
+	settings.destroy('config');
+	settings.destroy('page_sent');
+};
+
+/**
+ * Overload toString method to show the version.
+ * @return {string} The module's version.
+ */
+Tracking.prototype.toString = function() {
+	return 'oTracking version ' + version;
```
notlee, line 66:
```suggestion
	return 'oTracking version ' + this.version;
```

MIGRATION.md
```diff
@@ -13,6 +13,14 @@ The methods `setDomain` and `getDomain` have been removed, o-tracking only works
 
 The `api_key` configuration settings has been removed, it is no longer required as it is not used by Spoor or o-tracking in anyway.
 
+`tracking.link` has been removed, it's replacement is `tracking.click`.
+```diff
+-tracking.link.init();
++tracking.click.init();
+```
+
+Removed the deprecated `user.passport_id` from the data being tracked.
```
JakeChampion, line 22:
```suggestion
Removed the deprecated `user.passport_id` from the data being tracked.

Removed the deprecated `user.erights_id` from the data being tracked.
```

PR #270, opened 2020-07-22
Here lies Version 3 of o-tracking

JakeChampion:
We (Origami) are making changes to o-tracking to hopefully have it be in a better shape for a team to take ownership and maintenance of 🤞 

All work is being done is branches off of the `v3` branch and the pull-requests will be merged back into the `v3` branch.


wheresrhys:
Out of curiosity what, if any, breaking changes to the API are there?

MIGRATION.md
```diff
@@ -1,6 +1,51 @@
 
 # Migration
 
+## Migrating from v2 to v3
```
JakeChampion, line 4: @wheresrhys this covers all the breaking changes 👍

MIGRATION.md
```diff
@@ -1,6 +1,51 @@
...
+
+o-tracking now uses the Beacon API in all browsers which support it. In v2 this was an opt-in feature, now it is always enabled in browsers which support it. For browsers which do not support Beacon API o-tracking will revert to the XMLHTTPRequest API.
```
wheresrhys, line 6: ... there were good reasons not to do this in the past, namely that you can't detect network failures and attempt resending. At the time the trade off between
- beacon will always send, but may not succeed and we won't be able to retry
- XHTTP may not always send, but if the user comes back we get to retry

resulted in a decision to stick with XHTTP (IIRC when we switched to beacon on ft.com we saw a noticeable change in analytics). I'm not sure if that is still the right decision, but has this been discussed with the data platform?

(Incidentally, this is an excellent example of why it'd important to transfer ownership of o-tracking to a team that is more invested in the entire e2e tracking pipeline)

MIGRATION.md
```diff
@@ -1,6 +1,51 @@
...
+
+The `Tracking.prototype.utils` object has been removed.
```
wheresrhys, line 8: 😍

MIGRATION.md
JakeChampion: Data Platform have seen a more than double increase in submitted events by using the Beacon API versus the XMLHTTPRequest API.

wheresrhys: wow

JakeChampion: This is reference to turning on Beacon API and turning off XMLHTTPRequest API:
>We see page interaction events coming from 68% of article page views (not last page) compared to 46% and for the last page of visits we now have 46% compared to 19%.  So more than doubled the delivery on the last page and about a 50% improvement elsewhere.

:D

README.md
```diff
@@ -1,302 +1,293 @@
...
+# o-tracking
 
-Include in your product to send tracking requests to the Spoor API.
+Include in your product to send tracking requests to the [Spoor API](https://spoor-docs.herokuapp.com/).
 
-- [The spoor ecosystem](#the-spoor-ecosystem)
 - [Usage](#usage)
-- [Events](#events)
-- [Parameters](#parameters)
+- [How To View The Data](#how-to-view-the-data)
```
notlee, line 6: Looks like this is missing some headings, are we linting it?

README.md
```diff
@@ -1,302 +1,293 @@
...
 
-## The Spoor ecosystem
-![ScreenShot](https://rawgit.com/Financial-Times/o-tracking/master/resources/images/ngda-system-design.svg)
+## Usage
 
-[Spoor API docs](https://spoor-docs.herokuapp.com/)
+Check out [how to include Origami components in your project](https://origami.ft.com/docs/components/#including-origami-components-in-your-project) to get started with `o-tracking`.
 
-## Usage
+### How to view the data
```
notlee, line 16: I think this section might make more sense after instructions on how to setup o-tracking

README.md
```diff
@@ -1,302 +1,293 @@
...
+
+Once you have sent data into Spoor, you can view the data via Chartio.
+The FT has a Slack channel named #chartio_q_a where anyone can get help with Chartio.
+If you have never used Chartio before, you will need to [request access via this form](https://docs.google.com/forms/d/e/1FAIpQLSc1z1hTtnhe4KSXuZ220Zf53wTm0ucYVf2B3jYY5ZSyNW2VSw/viewform).
+To sign in to Chartio:
+- Open the [FT Chartio](https://chartio.com/financialtimes/) in your browser.
+- Press the "Sign in with Google" button and select your FT Google Account.
+
+### How to track extra click data using `data-trackable` attributes
```
notlee, line 25: I think this section might make more sense after basic setup instructions

README.md
```diff
@@ -1,302 +1,293 @@
...
-			};
-		}
-	}
 
-	oTracking.view.init(opts);
-	```
+#### Events
 
-Events are decorated with config values you pass in via `init()` or `updateConfig()` if they are part of `context`, `user`, or `device` objects. Values passed in as `context` to individual events will override context values from init.
+Instead of calling the `page` and `event` methods `o-tracking` can be configured to fire events based on the custom DOM Events `oTracking.page` and `oTracking.event`.
```
notlee, line 240: it would be good to expand here: why would I choose one approach over the other? Would it make sense to do both?

README.md
```diff
@@ -1,302 +1,293 @@
...
+// Now we can send custom events and o-tracking will track the data within them.
+const customData = { category: 'video', action: 'play', id: '512346789', pos: '10' };
+document.body.dispatchEvent(new CustomEvent('oTracking.event', { detail: customData}, bubbles: true}));
 ```
 
-### Event
+## How to view the data
 
-__Important__: To decide how to name the category and action fields, consider this sentence (square brackets denote that part is optional):
+Onec you have sent data into Spoor, you can view the data via Chartio.
```
notlee, line 268: Ah I think this is duplicated above, but does make more sense down here

notlee (commented):
I haven't reviewed the code since we've been reviewing smaller PRs which make up v3, and the changes are great 😄, but do have some readme feedback

README.md
JakeChampion: Fixed in https://github.com/Financial-Times/o-tracking/pull/309

README.md
JakeChampion: Fixed in #309

README.md
JakeChampion: Fixed in #309

README.md
JakeChampion: It's something I've only seen the FT App use but I can't see anything which fires the 'oTracking.page' event -- https://github.com/Financial-Times/ft-app/blob/83e4c16b34f79fe2923c9361f2e315cb2bc80695/lib/javascript/tracking/events.js#L11-L20

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to af1a2d177acd353456a3aab32608ecfa954e6706
js: 4.01kb increase (1.41kb/gzip)

notlee:
This looks great to me :D Are we ready to release or waiting on [this pr](https://github.com/Financial-Times/o-tracking/pull/310)?

JakeChampion:
@notlee we are ready to release, we won't wait on that PR 👍

notlee (approved):

PR #272, opened 2020-07-22
Remove documentation about deprecated user.erights_id tracking data

JakeChampion:
erights_id is not generated internally by o-tracking and is not used in Spoor at all. The only reference left in o-tracking is these two documentation files


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 85619c0f8032762e8144dca970fe603bc81c89aa
js: 5.15kb increase (1.62kb/gzip)

PR #275, opened 2020-07-22
Remove the named exports and only have the default export which is the singleton Tracking instance

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 564b9bd0052966d0753899322f9001c31bf73dcd
js: 4.96kb increase (1.55kb/gzip)

PR #276, opened 2020-07-22
Remove tracking.prototype.utils from the public API

JakeChampion:
These utils are not required for use outside of o-tracking's codebase


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 12cbd0802cf8f41b571815707fc77d96b599bae5
js: 5.13kb increase (1.61kb/gzip)

PR #277, opened 2020-07-23
Improve/fix the JSDoc types being declared across the codebase

JakeChampion:
I turned on JSDoc type-checking in my editor and came across lots of errors which stemmed from incorrect JSDoc comments. This pull-requests fixes the JSDoc so that o-tracking has no more incorrect errors. Some of the changes had to be done to the code for the JSDoc type-inference to kick in, such as replacing a reduce function with a for-of loop.

### DIFF
--- a/src/javascript/events/click.js
+++ b/src/javascript/events/click.js
@@ -72,7 +72,7 @@ const handleClickEvent = eventData => (clickEvent, clickElement) => {
 /**
  * If there are any requests queued, attempts to send the next one
  * Otherwise, does nothing
- * @return {undefined}
+ * @return {void}
  */
 /*eslint-disable no-unused-vars*/
 const runQueue = _ => {
@@ -90,7 +90,7 @@ const runQueue = _ => {
  * @alias click#init
  * @param {String} category - The event category for clicks.
  * @param {String} elementsToTrack - A query selector string to select elements to track links on {@link https://github.com/ftlabs/ftdomdelegate#selector-string}.
- * @return {undefined}
+ * @return {void}
  */
 const init = (category, elementsToTrack) => {
 	elementsToTrack = elementsToTrack || 'a, button, input, [role="button"]'; // See https://github.com/ftlabs/ftdomdelegate#selector-string
--- a/src/javascript/events/click.js
+++ b/src/javascript/events/click.js
@@ -7,13 +7,6 @@ import getTrace from '../../libs/get-trace';
 
 let internalQueue;
 
-const eventPropertiesToCollect = [
-	"ctrlKey",
-	"altKey",
-	"shiftKey",
-	"metaKey",
-];
-
 // Trigger the event tracking
 const track = eventData => {
 	const firstDomPathToken = eventData.context.domPathTokens[0];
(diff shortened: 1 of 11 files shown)


src/javascript/tracking.js
```diff
@@ -145,10 +145,10 @@ Tracking.prototype.init = function(config) {
 
 	settings.set('page_sent', false);
 
-	const cookieDomain = config ? config.cookieDomain : false;
+	const cookieDomain = config ? config.cookieDomain : '';
 
 	// Set up the user from stored - may later be updated by config
-	user.init(false, cookieDomain);
+	user.init('', cookieDomain);
```
JakeChampion, line 151: `User.init` takes two strings, if the strings are empty it triggers the function to set some default value.

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to fb0ad4606356b5e3e28346a21ec644437f55b5f0
js: 4.93kb increase (1.47kb/gzip)

src/javascript/events/custom.js
```diff
@@ -1,6 +1,18 @@
 import Core from '../core';
 import utils from '../utils';
 
+/**
+ * @typedef {Object} TrackingEvent
+ * @property {EventDetail} detail - The custom o-tracking event details
+ *
+ * @typedef {Event & TrackingEvent} OTrackingEvent
```
JakeChampion, line 8: This is creating a type-definition named OTrackingEvent which is a combination of the Event type-definition and the TrackingEvent type-definition. This is a way in JSDoc to mimic inheritance.

src/javascript/events/click.js
```diff
@@ -35,19 +28,28 @@ const track = eventData => {
...
 const getEventProperties = event => {
-	const eventProperties = eventPropertiesToCollect.reduce((returnObject, property) => {
-		try {
-			if (event[property]) {returnObject[property] = utils.sanitise(event[property]);}
-		}
-		catch (e) {
-			// eslint-disable-next-line no-console
-			console.log(e);
+	const eventProperties = {};
+	for (const property of eventPropertiesToCollect) {
```
notlee, line 42: 🙏

src/javascript/events/custom.js
notlee: interesting

src/javascript/events/page-view.js
```diff
@@ -26,8 +26,8 @@ const defaultPageConfig = function () {
  * Make the page tracking request.
  *
  * @param {Object} config - Configuration object. If omitted, will use the defaults.
- * @param {Function} callback - Callback function. Called when request completed.
- * @return {undefined}
+ * @param {Function=} callback - Callback function. Called when request completed.
```
notlee, line 29: what does the `=` indicate? _yeah, I'm being being lazy and could DuckDuckEcosia it_

JakeChampion: optional

notlee: optional?

JakeChampion: yeah, the equals sign states that the callback parameter is an optional parameter

notlee: cool :) I've always defined the default value as `undefined` but this is much easier

JakeChampion: `undefined` is technically not correct because then jsdoc/typescript will read it as the argument is required but can be set to undefined. By using the equals sign we are telling jsdoc/typescript that the argument does not need to be passed in at all.

```js
/**
 * log things
 *
 * @param {string} a - string to log out first.
 * @param {string|undefined} b - string to log out second - optional.
 * @return {void}
 */
function log(a, b) {
	if (b) {
		console.log(a, b);
	} else {
		console.log(a);
	}
}

log('hello'); // JSDoc/TypeScript will say this is missing a required argument.
```


```js
/**
 * log things
 *
 * @param {string} a - string to log out first.
 * @param {string=} b - string to log out second - optional.
 * @return {void}
 */
function log(a, b) {
	if (b) {
		console.log(a, b);
	} else {
		console.log(a);
	}
}

log('hello'); // JSDoc/TypeScript will say this is fine because argument `b` is marked as optional
```

PR #278, opened 2020-07-23
Turn tracking class into a singleton

JakeChampion:
Tracking is only ever used as a singleton object, let's turn it into a singleton to ensure it can only ever be used that way

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -11,10 +11,10 @@ The `api_key` configuration settings has been removed, it is no longer required 
 
 The deprecated `user.passport_id` data property has been removed.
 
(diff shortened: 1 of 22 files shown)


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to db2778b46d9b373ac2683c62622f4962b63d4679
js: 3.93kb increase (1.34kb/gzip)

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 5622e4ccbd7b8a7c5e0951d002deb59b73d6d1a5
js: 3.88kb increase (1.35kb/gzip)

PR #279, opened 2020-07-24
Turn the tracking class into a singleton object as tracking is only meant to be used as a singleton

JakeChampion:
The Tracking class is only ever meant to be used as a singleton object, let's make the implementation enforce that it can only be used as a singleton and not have multiple tracking instances by mistake.

This should not change anything for users of o-tracking as the exported interface is the same.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,6 +1,4 @@
-import { Tracking } from './src/javascript/tracking';
-
-const tracking = new Tracking();
+import tracking from './src/javascript/tracking';
 
 function initialise() {
 	tracking.init();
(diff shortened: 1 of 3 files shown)


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to cc4efbe593f13436d5a22a5baac0760c34c34421
js: 4.59kb increase (1.45kb/gzip)

JakeChampion:
@notlee Done 👍

notlee:
double approved 👍

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 6c06ec1361a2909d4d34d95ca18a57c714fe35ce
js: 4.59kb increase (1.45kb/gzip)

PR #285, opened 2020-07-27
Breaking: Always use the Beacon API if it exists.

JakeChampion:
Previously we had the Beacon API be an opt-in system.

Every browser except for Internet Explorer now supports Beacon API (https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon#Browser_compatibility) so we should use it by default and fallback to XHR for Internet Explorer.

Beacon API was purpose built for tracking/analytics requests.

With the Beacon API, data is transmitted asynchronously when the User Agent has an opportunity to do so, without delaying the next navigation (If the browser was navigating to a new page).

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -3,6 +3,8 @@
 
 ## Migrating from v2 to v3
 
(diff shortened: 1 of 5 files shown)


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to f7d1c5871cd3ef5f329acd1b6f9a6bc781e1e8a7
js: 4.68kb increase (1.40kb/gzip)

PR #286, opened 2020-07-27
Refactor: Use Object.assign instead of a for-in loop inside getDeclarativeConfig function

JakeChampion:
Hopefully this is easier to understand than the for-in loop


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to c53261159431e46408afc78e10c34fc6f11595c4
js: 4.4kb increase (1.38kb/gzip)

PR #287, opened 2020-07-27
Fix: Make JSDoc return type for getDeclarativeConfig correct by stating it can also return false

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 43c2ae2741721e02bb6696d165eb003069bc0f8d
js: 4.45kb increase (1.38kb/gzip)

PR #293, opened 2020-08-06
Remove the cookie based storage fallback

JakeChampion:
All browsers o-tracking supports have localStorage, it does not require the cookie based fallback.
Even in incognito sessions localStorage works.

[browser support for localStorage](https://caniuse.com/#feat=mdn-api_window_localstorage)

### DIFF
--- a/src/javascript/core/store.js
+++ b/src/javascript/core/store.js
@@ -8,6 +8,7 @@ import utils from '../utils';
  * @param {string} name - The name of the store
  * @param {Object} config - Optional, config object for extra configuration
  * @param {String=} config.nameOverride - The internal name for the store
+ * @param {String=} config.domain - The domain that should be used to store cookies on
  */
 const Store = function (name, config = {}) {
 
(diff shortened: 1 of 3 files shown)


chrisbrownuk:
@JakeChampion How will we manage the transition from using cookies to only using localStorage?  What about setting the localStorage value with the value from the cookie if it exists?  You could then delete the cookie if desired.

If we do this we won't get a whole load of new anonymous users and we'll be able to track the full history of anonymous users who go on to subscribe.

JakeChampion:
@chrisbrownuk excellent idea!

JakeChampion:
@chrisbrownuk I added a cookie importer and tests to prove it is working 👍

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 8a63a2be4d3f030db3438a3b6cfd96435706b210
js: 4.33kb increase (1.32kb/gzip)

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 034e2dbd7c8596a4e195d16f24d83fe16cffc6c2
js: 4.33kb increase (1.32kb/gzip)

chrisbrownuk:
> 
> 
> @chrisbrownuk I added a cookie importer and tests to prove it is working 👍

@JakeChampion Great.  Thank you.

src/javascript/core/store.js
```diff
@@ -6,17 +6,19 @@ import utils from '../utils';
...
+const Store = function (name, config = {}) {
 
 	/**
 	 * Internal Storage key prefix.
 	 */
 	const keyPrefix = 'o-tracking';
 
 
-	if (utils.isUndefined(name)) {
+	if (typeof name !== 'string' || name === '') {
```
notlee, line 21: 👍

src/javascript/core/store.js
```diff
@@ -152,6 +110,29 @@ const Store = function (name, config) {
...
 	}
 
+	// If the user has previous data stored in the old cookie storage system
+	// import the data into the new storage system and remove them from the cookie.
+	const oldCookieStoreData = cookieLoad(this.storageKey);
+	if (oldCookieStoreData) {
+		try {
+			const data = JSON.parse(oldCookieStoreData);
+			if (this.data) {
+				Object.assign(this.data, data);
```
notlee, line 120: Do we want keys in the cookie to replace keys in local storage?

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 4443db9a9ea28298c1f3805ce75156d537ad7021
js: 4.33kb increase (1.32kb/gzip)

src/javascript/core/store.js
JakeChampion: Good question, I was thinking we did, yes. It's not often that both storage are used on the same device so I didn't think it would matter which one override the other.

PR #295, opened 2020-08-11
Remove the default export for Queue

JakeChampion:
This does not effect the external interface at all


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to b088a48094541dc6d3020cf73eeb82e580b40699
js: 4.3kb increase (1.32kb/gzip)

PR #297, opened 2020-08-12
update the comment about why the first call to `page` does not set a new root_id

JakeChampion:
I found the original comment hard to understand so I decided to have an attempt at rewriting it.

### DIFF
--- a/src/javascript/events/page-view.js
+++ b/src/javascript/events/page-view.js
@@ -2,7 +2,7 @@ import Core from '../core';
 import utils from '../utils';
 import settings from '../core/settings';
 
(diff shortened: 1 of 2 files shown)


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 0462818de892f018e369cf19ca035866894e94d8
js: 4.35kb increase (1.34kb/gzip)

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 3911cfc9cfe4d36c9e729d9d8955d4616308ef43
js: 4.35kb increase (1.34kb/gzip)

Issue #296, opened 2020-08-12
Bug report: multiple page view events being fired in quick succession


JakeChampion:
## What
We are still seeing the same o-tracking issue as is shown in this closed issue https://github.com/Financial-Times/o-tracking/issues/134.

In this case it is on an Edge browser with 24k page view events for the same page.

More data here: https://docs.google.com/spreadsheets/d/1xkLDD-2kqnUWsAkH1WebunKKOU6No5hv6glcj099dTs/edit#gid=0.

JakeChampion:
The root_id value is different for each page view event, if that's useful.  Normally multiple events on the same page have the same root_id.

JakeChampion:
Okay, I’ve been looking into o-tracking some more.
The root_id would only change if the page was reloaded, this is because the only place in o-tracking where setRootID is called is in the function which sends a page view event to Spoor and that function ensures it only calls setRootID
 once by using another variable to indicate setRootID has been called. https://github.com/Financial-Times/o-tracking/blob/26d867b2bf780b000523b7bebc5353d4b50cba38/src/javascript/events/page-view.js#L37-L45

JakeChampion:
The only way I can get multiple page view events to trigger on the alphaville page is by emitting a custom event on the page.
If you run this in the developer console on the page it will trigger the alphaville code to send a pageview event.
```js
document.dispatchEvent(new Event('o.DOMContentLoaded'))
```
So, my hypothesis is that we have an advert which emits that event thousands of times and that is why we don’t see this issue on every page.

JakeChampion:
fixed via #298

PR #298, opened 2020-08-20
 Do not send multiple page-view events if the new event is the exact same as the previously sent event

JakeChampion:
Fixes https://github.com/Financial-Times/o-tracking/issues/296

### DIFF
--- a/src/javascript/events/page-view.js
+++ b/src/javascript/events/page-view.js
@@ -1,5 +1,5 @@
 import Core from '../core';
-import utils from '../utils';
+import {merge, triggerPage, addEvent, isDeepEqual} from '../utils';
 import settings from '../core/settings';
 
 settings.set('page_has_already_been_viewed', false);
@@ -20,6 +20,7 @@ const defaultPageConfig = function () {
 	};
 };
 
+let previousPageConfig = {};
 /**
  * Make the page tracking request.
  *
@@ -28,7 +29,7 @@ const defaultPageConfig = function () {
  * @return {void}
  */
 function page(config, callback) {
-	config = utils.merge(defaultPageConfig(), {
+	config = merge(defaultPageConfig(), {
 		context: config
 	});
 
@@ -43,11 +44,17 @@ function page(config, callback) {
 		Core.setRootID();
 	}
 	settings.set('page_has_already_been_viewed', true);
-
-	Core.track(config, callback);
-
-	// Alert internally that a new page has been tracked - for single page apps for example.
-	utils.triggerPage();
+	if (isDeepEqual(previousPageConfig, config)) {
+		if (settings.get('config').test) {
+			// eslint-disable-next-line no-console
+			console.warn('A page event has already been sent for this page, refusing to send a duplicate page event.');
+		}
+	} else {
+		previousPageConfig = config;
+		Core.track(config, callback);
+		// Alert internally that a new page has been tracked - for single page apps for example.
+		triggerPage();
+	}
 }
 
 /**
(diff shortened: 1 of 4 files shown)


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to d2988e5fba32a750635158a71c5d7825f8477095
js: 5.28kb increase (1.61kb/gzip)

chrisbrownuk:
@JakeChampion Please could you clarify if this will only remove events where the context.id, and everything else, are the same.  It looks like that's the case to me.  The issue [#296](https://github.com/Financial-Times/o-tracking/issues/296) had lots of events in rapid succession but they had different context.id and context.root_id values.

JakeChampion:
> @JakeChampion Please could you clarify if this will only remove events where the context.id, and everything else, are the same. It looks like that's the case to me. 

Yes, that is the case.

>The issue [#296](https://github.com/Financial-Times/o-tracking/issues/296) had lots of events in rapid succession but they had different context.id and context.root_id values.

Hmm, that would require a bit more work to solve in o-tracking, @chrisbrownuk do you think we should also solve that situation in o-tracking?

chrisbrownuk:
Thanks for clarifying @JakeChampion .

We do need to find a way to stop the issue happening.  It would be better to understand the cause so that we can put a specific fix in place, but this has been an occasional but recurring problem for years now and we still haven't worked out for sure why it happens.

We will need to be very careful in removing page view events in o-tracking.  Removing ajacent events that only differ by context.id and context.root_id would remove valid manual page refreshes.  So we would definitely need to consider the difference in timestamp and set a suitable threshold between adjacent page view events.  (Also possibly consider the number generated, but that'll be of little use in real time, since it's too late once they've been sent - I guess we could at least cap the number.)

chrisbrownuk (approved):

JakeChampion:
> We will need to be very careful in removing page view events in o-tracking. Removing ajacent events that only differ by context.id and context.root_id would remove valid manual page refreshes.

If the page is refreshed, then the local variable in o-tracking which keeps track of the previously sent page event would also be refreshed 👍

chrisbrownuk:
> If the page is refreshed, then the local variable in o-tracking which keeps track of the previously sent page event would also be refreshed 👍

Ah yes.  True.

jaya-ananthram-ft:
Hi @chrisbrownuk @JakeChampion , Just for my understanding this is the root cause for the issue https://financialtimes.slack.com/archives/C1WF81SR3/p1585756523025700?thread_ts=1585737775.022200&cid=C1WF81SR3 right?

JakeChampion:
> Hi @chrisbrownuk @JakeChampion , Just for my understanding, this is the root cause of the issue [financialtimes.slack.com/archives/C1WF81SR3/p1585756523025700?thread_ts=1585737775.022200&cid=C1WF81SR3](https://financialtimes.slack.com/archives/C1WF81SR3/p1585756523025700?thread_ts=1585737775.022200&cid=C1WF81SR3), right?

This is not the root cause no, we still don't know the root cause but this will stop the events from being making it all the way into Spoor

JakeChampion:
@chrisbrownuk and @jaya-ananthram-ft, I have added some more code to ignore duplicate page-view events that only differ by the context.id and context.root_id properties 👍

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 08b3d496b4bc45193cff02e309af81ad640d00aa

chrisbrownuk:
@JakeChampion Great.  Looks good.

notlee (approved):

PR #302, opened 2020-08-20
Make sure all functions have JSDoc and that the correct types are being used

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 3c2bf8a5cbf66ac7381b0e7c41b6c7faf2b5a5b7

PR #303, opened 2020-08-24
Make most of the internal modules use named exports instead of default exports

JakeChampion:


notlee (approved):

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to 11e3a70f21217e98ed799a371ac35583df59af2f
js: 4.01kb increase (1.41kb/gzip)

PR #306, opened 2020-09-04
add extension to relative imports to make the code work directly in a browser

JakeChampion:
relative imports require the full path including the extension because browsers use the import name to make a network request and they do not implicitly add a .js file extension if one is missing

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+/demos/local
+/bower_components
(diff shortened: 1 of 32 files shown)


chee (approved):

origamiserviceuser:
o-tracking bundle size difference from 2.0.10 to ec2e3154d53ce4d4cc062e756641ee4b37b4a668

PR #326, opened 2021-01-25
Export an object with `init` method for `click` and `view`.

notlee:
So `oTracking.click` becomes `oTracking.click.init` and
`oTracking.view` becomes `oTracking.view.init`.

- The documentation refers to `oTracking.click.init` and
`oTracking.view.init`. `oTracking.click` and `oTracking.view` are
undocumented.
- `oTracking.page.init` already exists, each event type will align.
`oTracking.page` is a function for sending a page event where
`oTracking.page.init` is a method for firing page events 
automatically. In the future view/click could follow the same pattern
-- e.g. to track clicks on non-interactive elements

This could be considered a breaking change but I think we should
treat this as a bugfix since there are few users of v3 and the
methods being changed are not documented.

Thanks for your help, @taraojo!

### DIFF
--- a/src/javascript/tracking.js
+++ b/src/javascript/tracking.js
@@ -4,10 +4,10 @@ import {init as initSession} from './core/session.js';
 import {init as initSend} from './core/send.js';
 import {event} from './events/custom.js';
 import {page} from './events/page-view.js';
-import {init as initClick} from './events/click.js';
+import click from './events/click.js';
 import core from './core.js';
 import { merge, broadcast } from './utils.js';
-import {init as initView} from './events/component-view.js';
+import view from './events/component-view.js';
 
 const initEvent = event.init;
 const initPage = page.init;
@@ -193,10 +193,10 @@ const tracking = {
 	destroy,
 	toString,
 	init,
-	click: initClick,
+	click,
 	event,
 	page,
-	view: initView,
+	view,
 	getRootID: core.getRootID
 };
 
--- a/src/javascript/tracking.js
+++ b/src/javascript/tracking.js
@@ -4,10 +4,10 @@ import {init as initSession} from './core/session.js';
 import {init as initSend} from './core/send.js';
 import {event} from './events/custom.js';
 import {page} from './events/page-view.js';
-import click from './events/click.js';
+import {click} from './events/click.js';
 import core from './core.js';
 import { merge, broadcast } from './utils.js';
-import view from './events/component-view.js';
+import {view} from './events/component-view.js';
 
 const initEvent = event.init;
 const initPage = page.init;
--- a/src/javascript/events/click.js
+++ b/src/javascript/events/click.js
@@ -118,6 +118,8 @@ const init = (category, elementsToTrack) => {
 	onPage(runQueue);
 };
 
-export {
+const click = {
 	init
 };
+
+export { click };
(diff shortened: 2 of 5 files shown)


notlee:
Also an issue to improve o-tracking test coverage 🙈 
https://github.com/Financial-Times/o-tracking/issues/327

origamiserviceuser:
o-tracking bundle size difference from 3.0.1 to e8fa7618e86cf570b073c1ca18bf74098d882618

src/javascript/tracking.js
```diff
@@ -4,10 +4,10 @@ import {init as initSession} from './core/session.js';
 import {init as initSend} from './core/send.js';
 import {event} from './events/custom.js';
 import {page} from './events/page-view.js';
-import {init as initClick} from './events/click.js';
+import {click} from './events/click.js';
```
taraojo, line 7: Thanks for looking into this @notlee 🙂 It looks like `'./events/click.js'` only has one named export `init`, so I'm not sure this doing what is expected 🤔

src/javascript/tracking.js
```diff
@@ -193,10 +193,10 @@ const tracking = {
 	destroy,
 	toString,
 	init,
-	click: initClick,
+	click,
```
taraojo, line 196: Would it be too odd to make this `click: { initClick }` or we could add a named export in the `click.js` file which returns an object?

notlee: I think `click.js` already returns a named exports `export {
	init
};`

origamiserviceuser:
o-tracking bundle size difference from 3.0.1 to 0e9d5a861b4c5b35e3fea2536de207bc43098cca

src/javascript/tracking.js
```diff
@@ -4,10 +4,10 @@ import {init as initSession} from './core/session.js';
 import {init as initSend} from './core/send.js';
 import {event} from './events/custom.js';
 import {page} from './events/page-view.js';
-import {init as initClick} from './events/click.js';
+import click from './events/click.js';
```
JakeChampion, line 7: I don't think this works. The `events/click.js` file exports only an `init`. This is currently importing something which does not exist.

src/javascript/tracking.js
taraojo: `init` in this case isn't an object, but the export of one item. If we were to make it a default then it would be an object, so could change this back to `export default { init };` and that would work with the other change -> `import click from './events/click.js';`

Or as a named export, maybe this would work, but it's not too clear

```js
const initClick = { init };
export { initClick as init };
```

notlee: ah my bad, I read that as a default export :)

src/javascript/tracking.js
notlee: ah my bad, I read the export as a default export :)

src/javascript/tracking.js
notlee: I've updated this to be more like `/events/page-view.js`. An object is returned with an `init` method

JakeChampion (approved):

taraojo (approved):
🎉

origamiserviceuser:
o-tracking bundle size difference from 3.0.1 to 1b6b48491b93dc808cbd914ae8806607947b09f7

origamiserviceuser:
:tada: This PR is included in version v3.0.2 :tada:

PR #336, opened 2021-04-01
Fix bug where loading spoor-id from the cookie would silently fail

JakeChampion:
spoor-id is stored directly as a string and not an json object, which meant json.parse would throw an error and we would not be migrating the old spoor-id value to localstorage and instead would mint a new spoor-id, which is not what we want to do.

### DIFF
--- a/src/javascript/core/store.js
+++ b/src/javascript/core/store.js
@@ -116,14 +116,20 @@ const Store = function (name, config = {}) {
 	const oldCookieStoreData = cookieLoad(this.storageKey);
 	if (oldCookieStoreData) {
 		try {
-			const data = JSON.parse(oldCookieStoreData);
-			if (this.data) {
-				Object.assign(this.data, data);
+			if (this.storageKey === 'spoor-id') {
+				// spoor-id is stored directly as a string and not as an object
+				this.data = oldCookieStoreData;
+				cookieRemove('spoor-id');
 			} else {
-				this.data = data;
-			}
-			for (const name of Object.keys(data)) {
-				cookieRemove(name);
+				const data = JSON.parse(oldCookieStoreData);
+				if (this.data) {
+					Object.assign(this.data, data);
+				} else {
+					this.data = data;
+				}
+				for (const name of Object.keys(data)) {
+					cookieRemove(name);
+				}
 			}
 		} catch (error) {
 			broadcast('oErrors', 'log', {
--- a/src/javascript/core/store.js
+++ b/src/javascript/core/store.js
@@ -119,7 +119,6 @@ const Store = function (name, config = {}) {
 			if (this.storageKey === 'spoor-id') {
 				// spoor-id is stored directly as a string and not as an object
 				this.data = oldCookieStoreData;
-				cookieRemove('spoor-id');
 			} else {
 				const data = JSON.parse(oldCookieStoreData);
 				if (this.data) {
@@ -128,7 +127,9 @@ const Store = function (name, config = {}) {
 					this.data = data;
 				}
 				for (const name of Object.keys(data)) {
-					cookieRemove(name);
+					if (name !== 'spoor-id') {
+						cookieRemove(name);
+					}
 				}
 			}
 		} catch (error) {
--- a/src/javascript/core/store.js
+++ b/src/javascript/core/store.js
@@ -53,6 +53,10 @@ const Store = function (name, config = {}) {
 			return window.localStorage.setItem(name, value);
 		},
 		remove: function (name) {
+			// We attempt to remove the item from the old cookie storage
+			// because otherwise the next time o-tracking is initialised for
+			// this user, it will import the old value which was meant to
+			// have been removed.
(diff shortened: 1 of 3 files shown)


src/javascript/core/store.js
```diff
@@ -116,11 +116,16 @@ const Store = function (name, config = {}) {
 	const oldCookieStoreData = cookieLoad(this.storageKey);
 	if (oldCookieStoreData) {
 		try {
-			const data = JSON.parse(oldCookieStoreData);
-			if (this.data) {
-				Object.assign(this.data, data);
+			if (this.storageKey === 'spoor-id') {
+				// spoor-id is stored directly as a string and not as an object
+				this.data = oldCookieStoreData
```
notlee, line 121:
```suggestion
				this.data = oldCookieStoreData;
```

notlee (approved):

src/javascript/core/store.js
```diff
@@ -116,14 +116,20 @@ const Store = function (name, config = {}) {
 	const oldCookieStoreData = cookieLoad(this.storageKey);
 	if (oldCookieStoreData) {
 		try {
-			const data = JSON.parse(oldCookieStoreData);
-			if (this.data) {
-				Object.assign(this.data, data);
+			if (this.storageKey === 'spoor-id') {
+				// spoor-id is stored directly as a string and not as an object
+				this.data = oldCookieStoreData;
+				cookieRemove('spoor-id');
```
Lily2point0, line 122: I can see the cookie removal being an issue for id consistency.
Scenario: user was on ft.com and has cookie set. user is put in the beta homepage cohort. local storage is set, cookie is removed. user clicks on an article from the beta homepage: a new cookie is set, creating another id mismatch.

Lily2point0 (approved):

origamiserviceuser:
o-tracking bundle size difference from 3.0.3 to 5117614f778630b4bc377e17b1921b7799eb0a4e
No bundle size differences found.

src/javascript/core/store.js
JakeChampion: oh wow, yes, great catch Lily!

Lily2point0:
I clicked 'approve' by force of habit, I actually had a comment first but I don't know how to remove my approval 😅

Lily2point0 (approved):
Approved (for good this time)

notlee:
Flying brain dump before leaving for lunch: If the cookie is not removed won't any data set to local storage be ignored on future visits? Maybe the content of the cookie should be emptied except ids and the check be based on that?

JakeChampion:
> Flying brain dump before leaving for lunch: If the cookie is not removed won't any data set to local storage be ignored on future visits? Maybe the content of the cookie should be emptied except ids and the check be based on that?

Good idea, I'll implement that 👍 

Just for refence, I check o-tracking v2 and there are 3 stores being used:
- A [store for the queue](https://github.com/Financial-Times/o-tracking/blob/a7e9b77bdcee9ecd4b98f8b8e3e13bacd255b66e/src/javascript/core/queue.js#L31), which will use [localStorage by default](https://github.com/Financial-Times/o-tracking/blob/a7e9b77bdcee9ecd4b98f8b8e3e13bacd255b66e/src/javascript/core/store.js#L50-L62) and fallback to cookies if localStorage does not exist 
- A [store for the session](https://github.com/Financial-Times/o-tracking/blob/a7e9b77bdcee9ecd4b98f8b8e3e13bacd255b66e/src/javascript/core/session.js#L5-L9), which will use [localStorage by default](https://github.com/Financial-Times/o-tracking/blob/a7e9b77bdcee9ecd4b98f8b8e3e13bacd255b66e/src/javascript/core/store.js#L50-L62) and fallback to cookies if localStorage does not exist 
- A [store for the user](https://github.com/Financial-Times/o-tracking/blob/a7e9b77bdcee9ecd4b98f8b8e3e13bacd255b66e/src/javascript/core/user.js#L15-L28), which always uses cookies and never localStorage

src/javascript/core/store.js
```diff
@@ -53,6 +53,7 @@ const Store = function (name, config = {}) {
 			return window.localStorage.setItem(name, value);
 		},
 		remove: function (name) {
+			cookieRemove(name);
```
JakeChampion, line 56:
```suggestion
			// We attempt to remove the item from the old cookie storage
			// because otherwise the next time o-tracking is initialised for
			// this user, it will import the old value which was meant to 
			// have been removed.
			cookieRemove(name);
```

origamiserviceuser:
o-tracking bundle size difference from 3.0.3 to a5b74598a330c49b0f3f3f0f246b95c1b6ab1a59
No bundle size differences found.

origamiserviceuser:
o-tracking bundle size difference from 3.0.3 to edf82b0a93e4c6ccd21ebfd193cb781e44534c13
No bundle size differences found.

origamiserviceuser:
:tada: This PR is included in version v3.0.4 :tada:

PR #337, opened 2021-04-22
convert to origami 2.0 component

JakeChampion:


test/main.test.js
```diff
@@ -46,7 +46,7 @@ describe('main', function () {
 		oTracking.destroy();
 	});
 
-	describe('a second instance of o-tracking exists in a separate browser context (e.g. new tab)', function () {
+	describe.skip('a second instance of o-tracking exists in a separate browser context (e.g. new tab)', function () {
```
JakeChampion, line 49: @notlee, I spent a while looking into why this test fails.
It is because the file being served by karma for `'/base/test/o-tracking-test.js'` is an empty file.
I'm not sure how to get it working, which is why I have skipped it here.

test/events/page.test.js
```diff
@@ -25,7 +25,7 @@ describe('page', function () {
 
 	after(function () {
 		new Queue('requests').replace([]); // Empty the queue as PhantomJS doesn't always start fresh.
-		destroy('config'); // Empty settings.
```
JakeChampion, line 28: This was a bug that never got hit in the previous test runner because this was always the last test to be executed.

The bug is that `destroy` will delete the config property when what we want to do is reset the config property to it's default value, which is an empty object.

Our new test runner hit this bug, which is why it was spotted now, I'll backport this to the current version of o-tracking 👍

chee (approved):
great work on this one! seemed hard

origamiserviceuser:
:tada: This PR is included in version v4.0.0 :tada:

PR #339, opened 2021-05-07
We don't want to use dependabot to update dependencies anymore

JakeChampion:


origamiserviceuser:
o-tracking bundle size difference from 3.1.1 to 6224cadc1be858438d71b48b8b0d7e72e595c3ea

PR #341, opened 2021-06-23
Bug: reset the config to it's default value instead of deleting it

JakeChampion:
This is a bug that doesn't currently get triggered by the test runner, likely because it is the last test file to be executed, but that could change whenever karma and/or mocha is updated.

The bug is that destroy will delete the config property when what we want to do is reset the config property to it's default value, which is an empty object.

Our new test runner in obt v11 hit this bug, which is why it was spotted now

### DIFF
--- a/test/events/page.test.js
+++ b/test/events/page.test.js
@@ -3,7 +3,7 @@
 
 import '../setup.js';
 
-import {destroy} from '../../src/javascript/core/settings.js';
+import {destroy, set} from '../../src/javascript/core/settings.js';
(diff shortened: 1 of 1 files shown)


origamiserviceuser:
o-tracking bundle size difference from 3.1.1 to bd7c5a083dba53ac34dbdbcfa48a06682bd56b5d

chee (commented):
very cool

chee (approved):

origamiserviceuser:
:tada: This PR is included in version v3.1.2 :tada:

PR #344, opened 2021-07-06
Read the old clicks queue used in v2 of o-tracking and make sure all within the queue are sent to spoor

JakeChampion:
There's currently a bug within o-tracking v3 and v4 where it does not ensure the events in the 'clicks' queue used in o-tracking v2 are sent to spoor-api.

This is an issue because it means some events can end up never being sent. The scenario that occurs in is when a product (such as front-page) uses o-tracking v2 and a user of the product clicks a link which takes them to a different product (such as article) which uses o-tracking v3 and then the user never goes back to any products which use o-tracking v2.

The fix for this is to have o-tracking read from the 'clicks' queue which v2 uses and send all the events stored in that queue off to spoor and remove them from the queue.

Note: We should backport this to v3 as well because otherwise users on v3 could end up losing this data as well

### DIFF
--- a/src/javascript/events/click.js
+++ b/src/javascript/events/click.js
@@ -3,6 +3,7 @@ import core from '../core.js';
 import {sanitise, assignIfUndefined, merge } from '../utils.js';
 import {get as getSetting} from '../core/settings.js';
 import {getTrace} from '../../libs/get-trace.js';
+import { Queue } from '../core/queue.js';
 
 let delegate;
 
@@ -72,7 +73,18 @@ const init = (category, elementsToTrack) => {
 	// Activate the click event listener
 	delegate = delegate || new Delegate(document.body);
 	delegate.on('click', elementsToTrack, handleClickEvent(eventData), true);
+
+	// Fire all click events that were stored in the old queue used by o-tracking v2
+	const clickQueue = new Queue('clicks');
+	sendAllEventsFromQueue(clickQueue);
 };
+
+function sendAllEventsFromQueue(queue) {
+	const nextLink = queue.shift();
+	if (nextLink) {
+		core.track(nextLink, () => sendAllEventsFromQueue(queue));
+	}
+}
 
 const click = {
 	init
--- a/test/events/click.test.js
+++ b/test/events/click.test.js
@@ -59,7 +59,7 @@ describe('click', function () {
 			// Add the click event to the old 'clicks' queue which o-tracking v2 uses
 			new Queue('clicks').replace([clickEventStoredInQueue]);
 		});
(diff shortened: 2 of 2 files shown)


src/javascript/events/click.js
```diff
@@ -72,8 +73,20 @@ const init = (category, elementsToTrack) => {
...
 	delegate.on('click', elementsToTrack, handleClickEvent(eventData), true);
+
+	// Fire all click events that were stored in the old queue used by o-tracking v2
+	const clickQueue = new Queue('clicks');
+	sendAllEventsFromQueue(clickQueue);
 };
 
+function sendAllEventsFromQueue(queue) {
+	const nextLink = queue.shift();
+	console.log({nextLink});
```
notlee, line 84:
```suggestion
```

chrisbrownuk (approved):
Thanks @JakeChampion

test/events/click.test.js
```diff
@@ -41,6 +41,45 @@ describe('click', function () {
...
+			}
+		};
+		beforeEach(function() {
+			// Add the click event to the old 'clicks' queue which o-tracking v2 uses
+			new Queue('clicks').replace([clickEventStoredInQueue]);
+		});
+		beforeEach(function() {
+			// Remove the events from the old 'clicks' queue
+			new Queue('clicks').replace([]);
+		});
```
jkerr321, line 65: I am a confused about why there are two `beforeEach` functions here and why the second doesn't 'overwrite' the output of the first?

JakeChampion: This was meant to be an afterEach, I'll update this now 👍

test/events/click.test.js
```diff
@@ -41,6 +41,45 @@ describe('click', function () {
...
+				},
+				"action": "click",
+				"category": "cta"
+			}
+		};
+		beforeEach(function() {
+			// Add the click event to the old 'clicks' queue which o-tracking v2 uses
+			new Queue('clicks').replace([clickEventStoredInQueue]);
+		});
+		beforeEach(function() {
```
JakeChampion, line 62:
```suggestion
		afterEach(function() {
```

jkerr321 (approved):

test/events/click.test.js
```diff
@@ -67,6 +106,7 @@ describe('click', function () {
 
 		setTimeout(() => {
 			try {
+				console.log(core.track)
```
JakeChampion, line 109:
```suggestion
```

origamiserviceuser:
:tada: This PR is included in version v4.1.1 :tada:

PR #345, opened 2021-08-02
fix the name of the optional feature navigator.sendBeacon

JakeChampion:
This matches the name of the feature within polyfill.io, we have the name match polyfill.io so that the origami-build-tools can include a call to polyfill.io when running the tests on our browser support matrix

### DIFF
--- a/origami.json
+++ b/origami.json
@@ -19,7 +19,7 @@
 		],
 		"optional": [
 			"Promise",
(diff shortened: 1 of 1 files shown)


chee (approved):

origamiserviceuser:
:tada: This PR is included in version v4.1.2 :tada:

PR #348, opened 2022-03-25
feat: add ability to force o-tracking to use a queue

JakeChampion:
This is a backport of https://github.com/Financial-Times/origami/pull/680

The reason for the backport is to allow the FT App to upgrade from v1 to v3 of o-tracking, which is the latest version on both npm and bower

---

By default o-tracking uses the browser Beacon API to send events to Spoor and the Beacon API only works when the device is online and o-tracking can not queue those events.

If wanting to queue events (to retry them if they failed to send) and/or wanting to record events when offline, set `queue` to `true` during the intialisation of `o-tracking` like so:

```js
import oTracking from 'o-tracking';

const config = {
    queue: true, // Make sure o-tracking stores the events on a local queue (this means o-tracking can work when the device is offline)
    ...
};
oTracking.init(config);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,22 @@ Here is the corresponding tracking pixel setup:
 
 ### Tracking with JavaScript
 
+### Turning on offline support and queueing events
+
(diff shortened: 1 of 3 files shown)


CLTPayne (approved):
🎉

